PR #5, opened 2023-08-10
Write to qbe intermediate language

leewei05:
This pull request only has the minimal changes to write `main` function to qbe intermediate language.

I'm planning to implement stmt node in `ast.cpp`, so that we can introduce statements to our ast.


Lai-YT:
I've noticed that your fork appears to be slightly behind the main branch. ☺️ 
Could you kindly consider rebasing it prior to address any potential conflicts?
Your effort is much appreciated! 😀

ast.cpp
```diff
@@ -28,12 +32,24 @@ class ProgramNode : public AstNode {
   ProgramNode(std::vector<std::unique_ptr<ExprNode>>&& exprs)
       : exprs_{std::move(exprs)} {}
 
+  void CodeGen() {
+    /* qbe main */
+    output << "export function w $main() {" << std::endl;
+    output << "@start" << std::endl;
+  }
```
Lai-YT, line 39:
```suggestion
  void CodeGen() {
    /* qbe main */
    output << "export function w $main() {" << std::endl;
    output << "@start" << std::endl;
    for (const auto& expr : exprs_) {
      expr->CodeGen();
    }
    output << " ret 0" << std::endl;
    output << "}";
  }

```
`CodeGen` should recursively call `CodeGen` on its sub-nodes, just like what `Dump` does.
`MainEnd` can then be inlined into `CodeGen`.

ast.cpp
```diff
@@ -42,6 +58,7 @@ class IntConstExprNode : public ExprNode {
  public:
   IntConstExprNode(int val) : val_{val} {}
 
+  void CodeGen() {}
```
Lai-YT, line 61:
```suggestion
  void CodeGen() override {}
```
Make the signature more explicit and can catch typo bugs, such as `void Codegen()`.

ast.cpp
```diff
@@ -56,6 +73,7 @@ class BinaryExprNode : public ExprNode {
   BinaryExprNode(std::unique_ptr<ExprNode> lhs, std::unique_ptr<ExprNode> rhs)
       : lhs_{std::move(lhs)}, rhs_{std::move(rhs)} {}
 
+  void CodeGen() {}
```
Lai-YT, line 76:
```suggestion
  void CodeGen() override {}
```

parser.y
```diff
@@ -23,24 +25,43 @@
 // Also with automove, smart pointers can be moved implicity without boilerplate std::move.
 // NOTE: can no longer reference a $x twice since it's moved in the first place.
 %define api.value.automove
+// This guarantee that headers do not conflict when included together
```
Lai-YT, line 28:
```suggestion
// This guarantees that headers do not conflict when included together.
```

Lai-YT (changes requested):

leewei05:
@Lai-YT Applied fixes! 🫡

Lai-YT (approved):

PR #6, opened 2023-08-10
Report error messages with std::cerr

Lai-YT:
This separates normal outputs from error messages so that redirections (>) can be better exploited.


leewei05 (approved):

leewei05:
LGTM 🚀

PR #7, opened 2023-08-10
Mark CodeGen as constant member function

Lai-YT:
Code generation is not a phase of transformation, which means it should not modify the syntax tree.
Optimizations, which perform the transformations, will be handled in a separate phase.


leewei05 (approved):
LGTM! 🚀

PR #8, opened 2023-08-11
Scan and parse identifier token

leewei05:
I think it would be handy if we introduce identifier before implementing statements.

### DIFF
--- a/ast.cpp
+++ b/ast.cpp
@@ -55,6 +55,18 @@ class ProgramNode : public AstNode {
   std::vector<std::unique_ptr<ExprNode>> exprs_;
 };
 
+class IdExprNode : public ExprNode {
+ public:
+  IdExprNode(const std::string& id) : id_{id} {}
+  void CodeGen() const override {}
+  void Dump(int pad) const override {
+    std::cout << Pad(pad) << id_ << std::endl;
+  }
+
+ protected:
+  std::string id_;
+};
+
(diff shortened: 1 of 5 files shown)


lexer.l
```diff
@@ -22,6 +22,9 @@
 #define LINENO (yylineno - 1)
 %}
 
+/* C99 7.1.3: All identifiers that begin with an underscore and either an uppercase letter or another
```
Lai-YT, line 25: I suggest that we have our line length under 80 characters 😀 
See [Google C++ Style Guide: Line Length](https://google.github.io/styleguide/cppguide.html#Line_Length) for the details!

ast.cpp
```diff
@@ -55,6 +55,18 @@ class ProgramNode : public AstNode {
   std::vector<std::unique_ptr<ExprNode>> exprs_;
 };
 
+class IdExprNode : public ExprNode {
+ public:
+  IdExprNode(std::string id) : id_{id} {}
```
Lai-YT, line 60:
```suggestion
  IdExprNode(const std::string& id) : id_{id} {}
```

String copying is considered an expensive operation and is unnecessary in this context.
Prefer passing `std::string`s as const-reference over value.

Lai-YT (changes requested):

leewei05:
@Lai-YT updated!

Lai-YT (approved):
LGTM 🎉

PR #11, opened 2023-08-12
Implement statement node in ast

leewei05:
Implement null and return statement.


parser.y
```diff
@@ -55,14 +57,21 @@ entry: main_func {
...
+    auto stmts = $1;
+    stmts.push_back($2);
+    $$ = std::move(stmts);
+  }
+  | epsilon { $$ = std::vector<std::unique_ptr<StmtNode>>{}; }
   ;
 
+stmt: ';' { $$ = std::make_unique<NullStmtNode>(); }
+    | RETURN exprs ';' { $$ = std::make_unique<ReturnStmtNode>($2); }
+    ;
```
Lai-YT, line 73: How about adding support for `expr ';'` together?
It's a valid statement, and we can keep most of our `arithmetic.c` by only adding `;`s.
Note: I'm amused by the fact that the present arithmetic.c is unrelated to "arithmetic" 😆

Lai-YT (changes requested):

leewei05:
@Lai-YT Applied fixes, removed unused parser rules and code.

Lai-YT (approved):
LGTM 🎉

PR #16, opened 2023-08-19
Test lexer & parser with Turnt

Lai-YT:
An external Python tool, [Turnt](https://github.com/cucapra/turnt), is introduced to orchestrate the tests. It runs a command and checks whether the output matches the expectation. The _*{filename}.exp_ files contain the expected outputs corresponding to _*{filename}.c_.
To install, execute `pip install turnt`. You are recommended to use the [Virtual Environment](https://docs.python.org/3.10/tutorial/venv.html) :grinning: .

8 test cases and a new Makefile target, `test`, are added to test the functionality of the lexer and the parser.These tests utilize the `Dump` function within the AST. It's worth noting that the lexer isn't tested in isolation; its testing is integrated with the parser. This simplification is deliberate, as creating a separate test for the lexer would necessitate a manual invocation of `lex()`.

Feel free to suggest changes or additional test cases if you believe there are gaps in coverage!


leewei05 (approved):
LGTM 🚀 🎉

PR #17, opened 2023-08-24
Convert ast file to header file

Lai-YT:
The _ast.cpp_ file provides classes or functions that are meant to be includes by others, such as _parser.y_.
It should be a header file rather than an implementation file.


leewei05 (approved):
LGTM 🎸

PR #23, opened 2023-08-27
Check type on simple int program

Lai-YT:
## What's New?

- Introduce `ExprType` to represent the type of identifiers (`DeclNode`) and expressions (`ExprNode`)
- Introduce `Environment` that employs stacked `SymbolTable`s to to express and manage the notion of scopes
- Add `CheckType` pass to resolve the type of nodes and check their compatibility
- Show types with `Dump`

## How to Review?

1. I suggest starting by reviewing the test cases to to gain insight into what we can achieve with the changes of this PR.
2. Subsequently, delve into the `CheckType` routine originating from the `ProgramNode`. Most node types feature straightforward implementations, often involving only recursive calls. Notable exceptions are `BlockStmtNode`, `DeclNode`, `ReturnStmtNode` and `IdExprNode`. They interact with the `Environment` and perform type checks. Ensure a solid grasp of their functionality.
3. Finally, explore the implementations of `SymbolTable` and `Environmemt`. It's advisable to avoid examining them prior to comprehending how `CheckType` interfaces with them, as It's always better to understand objects from their use cases.

Note that the current `ExprType` implementation is relatively basic, as it currently only accommodates the `int` type.

## Future Work

- Report messages on errors


env.hpp
leewei05, line 1: I think the name `env.hpp` is too vague, `scope.hpp` may be more direct?
Maybe we can rename the class to `Scope`.

Lai-YT: Great suggestion! Let me work on refining it :pen:

Given that `Environment` essentially represents a stack of scopes, I propose modifying the name to `ScopeStack` rather than just `Scope`. This adjustment more accurately conveys the concept of a collection of scopes organized in a stacked manner.

leewei05 (approved):
LGTM! 🎸 🧑‍🎤

PR #24, opened 2023-08-31
Move analysis and transformation outside parser

Lai-YT:
This adjustment ensures that the parser can stay focused on its primary responsibilities. Additionally, it helps minimize the potential for parsing failures, as errors related to analyses and transformations are separated from parsing errors.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1,9 +1,14 @@
 #include <fstream>
+#include <memory>
 
+#include "ast.hpp"
+#include "scope.hpp"
 #include "y.tab.h"
 
 /// @brief Where the generated code goes.
 std::ofstream output;
+/// @brief The root node of the program.
+auto program = std::unique_ptr<AstNode>{};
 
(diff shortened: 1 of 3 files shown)


main.cpp
```diff
@@ -14,7 +19,20 @@ int main(int argc, char** argv) {
   int ret = parser.parse();
 
   yylex_destroy();
+
+  // 1 if parsing failed because of invalid input;
+  // 2 if parsing failed due to memory exhaustion.
```
leewei05, line 24: Only found this in the manual [1]. Just curious where did you know the definition of return value.

> Method on parser: `int parse ()`
> Run the syntactic analysis, and return 0 on success, 1 otherwise. 

[1] https://www.gnu.org/software/bison/manual/html_node/C_002b_002b-Parser-Interface.html#index-parse-on-parser

leewei05 (approved):
LGTM! 🚀

PR #25, opened 2023-08-31
Rename Makefile variables to adhere to convention

Lai-YT:
This adjustment is crucial if you want to utilize implicit rules, as others are not recognized by the Makefile.
For further details, please refer to:
https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html


leewei05 (approved):
LGTM! 🦮

PR #31, opened 2023-09-10
Separate implementation from headers

Lai-YT:
This PR separates the implementation from headers, enabling incremental builds and promoting the decoupling of objects. Dependencies between files are now automatically managed using the `MMD` flag, and implementation files (e.g., `*.c`, `*.cpp`) are compiled utilizing [implicit rules](https://www.gnu.org/software/make/manual/html_node/Implicit-Rules.html).

Regarding the project layout, I propose that we adopt [The Pitchfork Layout (PFL)](https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs) as a recommended structure.
> Please note that I haven't delved into the details of PFL yet :face_in_clouds:. However, we can refer to it whenever we have questions or need guidance regarding the project's layout.


leewei05 (approved):
LGTM! 🚀

PR #32, opened 2023-09-11
Use unnamed namespace instead of static

Lai-YT:
Unnamed namespaces are more versatile in modern C++, as they support even class declarations.
For in-depth information, please consult the following resource: [Unnamed/anonymous namespaces vs. static functions](https://stackoverflow.com/questions/154469/unnamed-anonymous-namespaces-vs-static-functions)


leewei05 (approved):
LGTM!

PR #39, opened 2023-09-16
Check type with the Visitor pattern

Lai-YT:
## How to Review

1. Take a look at `main.cpp` to see how the `TypeChecker` is being used.
2. In `src/type_checker.cpp`, you can find the logic for type checking. Compare the differences with the original `CheckType` functions that are located in each corresponding AST node. You should discover that Visitors may have state: the `ScopeStack` is no longer being passed as an argument.  
3. Examine `include/visitor.hpp` and read the comments. You should discover that each AST node has a corresponding `Visit()` function, and there are two kinds of Visitors based on the constancy of the AST node being passed.
4. Finally, review `include/ast.hpp`. Each AST node has two `Accept()` functions to accept the corresponding kind of Visitor, and their implementations in `src/ast.cpp` are all identical and straightforward. You may want to refer to the concept of [Visitor and Double Dispatch](https://refactoring.guru/design-patterns/visitor-double-dispatch) to understand why we need these `Accept()` functions.

> **Note**
> 1. All AST nodes are now made public without removing the suffix from the names of those being affected. Additionally, the old `CheckType` functions are retained intentionally. Both of these decisions are deliberate, as changing them would necessitate numerous minor modifications, which would complicate the reviewing process. Keeping the old functions allows reviewers to compare the differences.
> 2. I prefer to treat AST nodes as data classes and centralize operations within visitors. In this context, exposing data members publicly is a reasonable choice. However, it's worth noting that in C++, constant-ness doesn't propagate down to pointers, which can affect the true constancy of objects. We may revisit the design implications of this decision in the future.

## Further Work

- Remove the suffix from public data members and member functions.
- Remove `CheckType()` functions.

### DIFF
--- a/include/ast.hpp
+++ b/include/ast.hpp
@@ -56,7 +56,6 @@ class DeclNode : public AstNode {
 
   void CheckType(ScopeStack& env) override;
 
- protected:
   std::string id_;
   ExprType type_;
   std::unique_ptr<ExprNode> init_;
@@ -78,7 +77,6 @@ class BlockStmtNode : public StmtNode {
 
   void CheckType(ScopeStack& env) override;
 
- protected:
   std::vector<std::unique_ptr<DeclNode>> decls_;
   std::vector<std::unique_ptr<StmtNode>> stmts_;
 };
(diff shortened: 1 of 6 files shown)


leewei05 (approved):
LGTM! 🚀 🚀 🚀

PR #40, opened 2023-09-20
Fix type checker calling deprecated type check

Lai-YT:
- Replace call to deprecated `TypeCheck`s with call to `Accept`
(fix #39 )
- Remove deprecated `TypeCheck`s

The `TypeChecker` encountered an issue where concrete binary expressions, such as `PlusExprNode`, were erroneously dispatched to the default implementation, resulting in no action being taken.
To resolve this, the `TypeChecker` has to implement the concrete binary expressions, explicitly dispatch them to the parent class `BinaryExprNode`.

### DIFF
--- a/include/ast.hpp
+++ b/include/ast.hpp
@@ -27,6 +27,7 @@ class AstNode {
 
 /// @note This is an abstract class.
 class StmtNode : public AstNode {
+ public:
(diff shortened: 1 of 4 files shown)


leewei05 (approved):
LGTM! 🌞

PR #42, opened 2023-09-24
Dump with Visitor pattern

Lai-YT:
This PR follows the same concept as #39, with the removal of deprecated `Dump` functions.
To improve code structure, I introduce a new object, `Indenter`, which handles indention details.

One potentially complex aspect is how we retrieve the `Op`s of `BinaryExprNode`s. This is accomplished using a hidden Visitor pattern. It's stateless and utilized only once in the `Visit` function of `BinaryExprNode`. As such, I've constructed it in-place rather than as a static global object or a data member of `AstDumpr`. For more details, please review the `OpGetter` class and its accompanying internal comments.


leewei05 (approved):
LGTM! 🔥

PR #43, opened 2023-09-30
CodeGen with Visitor pattern

Lai-YT:
- The `OpNameGetter` implementation closely follows the pattern set in `OpGetter` as established in #42.
- As part of this update, local numbers are no longer passed along with return values. Instead, we've introduced an auxiliary object responsible for recording these local numbers. This change eliminates the need for a dummy number, as now, only `ExprNode` generates temporaries with local numbers and maintains records of them.
- It's worth noting that this modification has transformed the Abstract Syntax Tree into a pure data structure. In subsequent commits, we may consider converting it into`struct`s to emphasize this evolution.

### DIFF
--- a/include/ast.hpp
+++ b/include/ast.hpp
@@ -16,7 +16,8 @@ class AstNode {
   virtual void Accept(NonModifyingVisitor&) const;
   virtual void Accept(ModifyingVisitor&);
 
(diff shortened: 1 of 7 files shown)


leewei05 (approved):
LGTM! 👍🏼

PR #44, opened 2023-10-03
Organize tests of type checking into named subfolder

Lai-YT:
Previously, tests of type checking were scattered throughout the root of the test folder, providing no clear indication of their purpose. By relocating them to a dedicated subfolder, we can now precisely identify the target of these tests through the folder's name.

> [!note]
> 1. We've chosen to name this subfolder `typecheck` in keeping with the naming convention used for `codegen`. However, we welcome any suggestions for improvement.
> 2. While these tests may also evaluate aspects of lexing and parsing, we have consolidated them to emphasize their primary focus on type checking, addressing the highest level of concern.


leewei05 (commented):
I'm a bit concerned that if we merged this PR, then we would have different test folders for frontend and backend. Yes, we would have a more detailed testing for both frontend and backend, but we would also need to create two source files in both `typecheck` and `codegen` directories. Maybe we can leave source test files in `test` root directory, and move `*.exp` to `typecheck` or `codegen`. By doing this, we would only need to create one source test file and have different expected test output for typecheck and codegen. Is this doable using turnt?

Lai-YT:
If we were to share the source test files, the tests would have to rely on identical test files. However, the concerns of tests at different phases are generally not going to be the same, so sharing them may not be feasible. Additionally, in turnt, a set of source test files shares a single configuration file. It seems like it doesn't support multiple configurations, although we could potentially achieve this by using a script to switch configurations.

Regarding the issue of duplicate `frontend` and `backend` folders across the structure, I don't think such duplication will pose a problem. Certain functionalities typically won't span across both frontend and backend. For instance, there's no type checking in the backend, and the input for code generation in the backend should be the IR rather than C, which is a fundamentally different test compared to the frontend QBE IR generation. Therefore, this should not be a significant concern.

leewei05:
> If we were to share the source test files, the tests would have to rely on identical test files. However, the concerns of tests at different phases are generally not going to be the same, so sharing them may not be feasible. Additionally, in turnt, a set of source test files shares a single configuration file. It seems like it doesn't support multiple configurations, although we could potentially achieve this by using a script to switch configurations.
> 
> Regarding the issue of duplicate `frontend` and `backend` folders across the structure, I don't think such duplication will pose a problem. Certain functionalities typically won't span across both frontend and backend. For instance, there's no type checking in the backend, and the input for code generation in the backend should be the IR rather than C, which is a fundamentally different test compared to the frontend QBE IR generation. Therefore, this should not be a significant concern.

Fair enough 👍🏼

leewei05 (approved):

PR #45, opened 2023-10-04
Convert AST nodes to struct & remove suffix from public data members

Lai-YT:
As discussed in issue #43, all AST nodes are meant to be pure data structures.
Additionally, in response to issue #39, suffixes have been removed.


leewei05 (approved):
LGTM! 🚀

PR #46, opened 2023-10-10
Rename ast dumper file name

leewei05:


Lai-YT:
> Checking CI lint

Simply reformat that will do the work.
The `#include "ast_dumper.hpp"` should be the first include. This coding convention is simply done by matching the file names between .cpp and .hpp, but since I've made a mistake on the file name of the .cpp file, the `#include "ast_dumper.hpp"` is treated as a normal include, thus not being the first include automatically.

Lai-YT (approved):
LGTM :tada:
Sorry for the typo :sob:

PR #47, opened 2023-10-13
Fix Clang warning and change C files to Cpp files

leewei05:
- Fix Clang mismatch warning

```
include/ast.hpp:14:1: warning: 'AstNode' defined as a struct here but previously declared as a class; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
struct AstNode {
^
include/visitor.hpp:9:1: note: did you mean struct here?
class AstNode;
^~~~~
struct
```

- Use cpp files instead of c files as suggested by Clang

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ YACC = bison
 # -d: generate header with default name
 YFLAGS = --verbose --debug -d
 
-# Note that lex.yy.c is excluded deliberately, as "lex.yy.c" is considered a
-# header file (it's included by "y.tab.c").
-OBJS := $(shell find . -name "*.cpp") y.tab.o
+# Note that lex.yy.cpp is excluded deliberately, as "lex.yy.cpp" is considered a
+# header file (it's included by "y.tab.cpp").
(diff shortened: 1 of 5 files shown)


Lai-YT:
Thank you for addressing this issue! :grinning: 
I didn't encounter any warnings during the compilation process. Could you please confirm whether you applied specific compiler flags or settings to trigger these warnings? Identifying and addressing such issues early in the development process can be quite helpful.

Lai-YT:
> @Lai-YT I didn't apply any specific flags to trigger these warnings. Maybe it's because I'm using Clang on macOS. 🤔 Are you using Clang or g++ on Ubuntu? Our compiler in our Makefile seems to be using g++, and it is linked to Clang on macOS. 😏

Haha, that's interesting to hear! It's good to know how macOS behaves in this case.
Time to test with both _clang++_ and _g++_!  :sunglasses:

Lai-YT (approved):
LGTM :love_you_gesture:

PR #48, opened 2023-10-13
Run tests with g++ and clang++

Lai-YT:
These two mainstream compilers may discover warnings that the other does not catch (an example in #47), so we'd like to run tests with both.

> [!note]
> I prefer to add the `-Werror` flag only when running tests in our CI environment.
This way, we don't have to fix all warnings during development at the first moment.
However, it appears that we cannot append _Makefile_ variables via the command line, as the [`override` directive](https://www.gnu.org/software/make/manual/html_node/Override-Directive.html) doesn't achieve what I have in mind.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,9 @@ jobs:
           scripts/check-format.sh
   test:
     runs-on: ubuntu-22.04
+    strategy:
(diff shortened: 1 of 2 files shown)


leewei05 (approved):
LGTM! 👍🏼

PR #49, opened 2023-11-27
Parse If statement

leewei05:
- Parse if statement
- Combine declarations and statements into one block
- Add simple if statement test case

### DIFF
--- a/include/ast.hpp
+++ b/include/ast.hpp
@@ -93,6 +93,18 @@ struct NullStmtNode : public StmtNode {
   virtual void Accept(ModifyingVisitor&) override;
 };
 
+struct IfStmtNode : public StmtNode {
+  IfStmtNode(std::unique_ptr<ExprNode> expr,
+             std::unique_ptr<BlockStmtNode> block)
+      : predicate{std::move(expr)}, body{std::move(block)} {}
+
+  virtual void Accept(NonModifyingVisitor&) const override;
+  virtual void Accept(ModifyingVisitor&) override;
+
+  std::unique_ptr<ExprNode> predicate;
+  std::unique_ptr<BlockStmtNode> body;
+};
+
 struct ReturnStmtNode : public StmtNode {
   ReturnStmtNode(std::unique_ptr<ExprNode> expr) : expr{std::move(expr)} {}
 
--- a/parser.y
+++ b/parser.y
@@ -109,7 +109,7 @@ stmts: stmts stmt {
 stmt: ';' { $$ = std::make_unique<NullStmtNode>(); }
     | RETURN expr ';' { $$ = std::make_unique<ReturnStmtNode>($2); }
     | expr ';' { $$ = std::make_unique<ExprStmtNode>($1); }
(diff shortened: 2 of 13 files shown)


parser.y
```diff
@@ -80,8 +75,13 @@ entry: main_func {
...
 
   /* TODO: mix declarations and statements in compound statement */
-main_func: INT MAIN '(' ')' '{' decls stmts '}' {
-    $$ = std::make_unique<BlockStmtNode>($6, $7);
+main_func: INT MAIN '(' ')' '{' block '}' {
+    $$ = $6;
+  }
+  ;
+
+block: decls stmts {
```
Lai-YT, line 83:
```suggestion
block: '{' decls stmts '}' {
```
In the C standard ISO/IEC 9899:1999, section 6.8.2 defines the syntax of a _compound statement_, also known as a _block_.

- $\text{compound-statement}$:
    **{** $\text{block-item-list}_\text{opt}$ **}**
- $\text{block-item-list}$:
    $\text{block-item}$
    $\text{block-item-list}$ $\text{block-item}$
- $\text{block-item}$:
    $\text{declaration}$
    $\text{statement}$

It's important to note that we are combining these 3 non-terminals together.
The key point is that the curly braces (`{`, `}`) are an integral part of the _block_.
This also simplifies references to the _block_ by eliminating the need to include the curly braces every time.
The _block_ always occurs with the curly braces.

include/ast.hpp
```diff
@@ -93,6 +93,18 @@ struct NullStmtNode : public StmtNode {
...
 
+struct IfStmtNode : public StmtNode {
+  IfStmtNode(std::unique_ptr<ExprNode> expr,
+             std::unique_ptr<BlockStmtNode> block)
+      : expr{std::move(expr)}, block{std::move(block)} {}
+
+  virtual void Accept(NonModifyingVisitor&) const override;
+  virtual void Accept(ModifyingVisitor&) override;
+
+  std::unique_ptr<ExprNode> expr;
```
Lai-YT, line 104:
```suggestion
  std::unique_ptr<ExprNode> predicate;
```
The expression should not only represent an expression but also have its type convertible to `bool` after evaluation.
I suggest that we change the name from `expr` to `predicate` to better reflect this characteristic.
Aslo `if_stmt->expr` can be a little bit confusing.

Lai-YT (changes requested):

include/ast.hpp
```diff
@@ -93,6 +93,18 @@ struct NullStmtNode : public StmtNode {
...
+  std::unique_ptr<BlockStmtNode> block;
```
Lai-YT, line 105:
```suggestion
  std::unique_ptr<BlockStmtNode> body;
```
Suggest naming based on semantics rather than types.
Additionally, for future _if statement_ that incorporates both _then_ and _else_ parts, using `then` and `else` for naming would be a considerable choice.

Lai-YT (changes requested):

PR #50, opened 2023-11-27
Implement if statement code generation

leewei05:
One important change is to move the `start` label from `BlockStmtNode` to `ProgramNode`. The reason is that `BlockStmtNode` cannot output the correct label to its own block because it doesn't know whether it is a if statement body or a function body. For instance, if we keep `start` at the `BlockStmtNode`, then it will have label `start` and label `@if.x` at the beginning of a if statement body. Thus, by moving the part where we create labels for each block to an upper level, each block can have its correct starting label.


src/qbe_ir_generator.cpp
```diff
@@ -203,27 +227,27 @@ class OpNameGetter::OpNameGetterImpl : public NonModifyingVisitor {
   }
 
   void Visit(const GreaterThanExprNode&) override {
-    op_name_ = "sgt";
+    op_name_ = "csgtw";
   }
 
   void Visit(const GreaterThanOrEqualToExprNode&) override {
-    op_name_ = "sge";
+    op_name_ = "csgew";
```
Lai-YT, line 234: What is the reasoning behind this update?
My assumption is that we missed the `c` prefix, making `sge` itself an invalid op.
Additionally, is the `w` suffix indicating integer ("word") type?
Am I correct in this interpretation?

Lai-YT (changes requested):

src/qbe_ir_generator.cpp
leewei05: Yep. The correct comparison instructions are in this section[1].

[1] https://c9x.me/compile/doc/il.html#Instructions-Index

Lai-YT: Understand. I suggest using the verb `Fix` instead of `Update` in the commit message to tell better that the original is ill-formed or invalid 👍 .

leewei05: Done! Great suggestion. 💯

src/qbe_ir_generator.cpp
```diff
@@ -22,11 +22,21 @@ int NextLocalNum() {
...
+
 /// @brief Returns the function-scope temporary with sigil (`%`).
 std::string PrefixSigil(int local_num) {
   return "%." + std::to_string(local_num);
 }
 
+/// @brief Returns a label with prefix (`@`).
+std::string PrefixLabel(const std::string& name, int local_num) {
+  return "@" + name + "." + std::to_string(local_num);
+}
```
Lai-YT, line 38:
```suggestion
std::string PrefixLabel(const std::string& name, int label_num) {
  return "@" + name + "." + std::to_string(label_num);
}
```

Lai-YT: The renaming seems to be missed!

Lai-YT (changes requested):
Once this small issue is addressed, I'll be ready to approve the PR 😊!

src/qbe_ir_generator.cpp
leewei05: Sorry! It's fixed now! 🙏🏼

Lai-YT (approved):
LGTM 🚀

PR #51, opened 2023-11-28
Implement If else statement code parsing and code generation

leewei05:


include/ast.hpp
```diff
@@ -95,14 +95,18 @@ struct NullStmtNode : public StmtNode {
 
 struct IfStmtNode : public StmtNode {
   IfStmtNode(std::unique_ptr<ExprNode> expr,
-             std::unique_ptr<BlockStmtNode> block)
-      : predicate{std::move(expr)}, body{std::move(block)} {}
+             std::unique_ptr<BlockStmtNode> first_body,
+             std::unique_ptr<BlockStmtNode> second_body)
+      : predicate{std::move(expr)},
+        then{std::move(first_body)},
+        or_else{std::move(second_body)} {}
```
Lai-YT, line 102:
```suggestion
             std::unique_ptr<BlockStmtNode> then,
             std::unique_ptr<BlockStmtNode> or_else = {})
      : predicate{std::move(expr)},
        then{std::move(then)},
        or_else{std::move(or_else)} {}
```

Here are 2 suggested changes:
1. For better clarity, I propose renaming the parameters. The function signature is crucial documentation, and expressive parameter names help convey their intended use. I don't personally use the terms "first body" and "second body" to denote the "then" and "else" parts. These names might be a bit confusing.
Just a quick note to let you know: We are using the [constructor initializer list](https://en.cppreference.com/w/cpp/language/constructor), and the compiler can distinguish between data members and parameters even if they share the same name. If initialization is done in the body of the constructor, the data members would have to be accessed through the `this` pointer (e.g., `this->then`).
2. To align with a consistent style, suggest making the "else" part default to `nullptr`. This is similar to the approach taken in the `DeclNode`:
    ```c++
      DeclNode(const std::string& id, ExprType decl_type,
           std::unique_ptr<ExprNode> init = {})
      : id{id}, type{decl_type}, init{std::move(init)} {}
    ```
    With this change, the _action_ of the production rule can be updated as follows:
    ```yacc
    IF '(' expr ')' block { $$ = std::make_unique<IfStmtNode>($3, $5); }
    ```

Lai-YT (changes requested):

include/ast.hpp
leewei05: Thanks for pointing out! I agree that this is a better solution for improving readability. 😄

src/qbe_ir_generator.cpp
```diff
@@ -124,14 +124,28 @@ void QbeIrGenerator::Visit(const IfStmtNode& if_stmt) {
   if_stmt.predicate->Accept(*this);
   int predicate_num = num_recorder.NumOfPrevExpr();
   int label_num = NextLabelNum();
-  std::string start_label = PrefixLabel("if", label_num);
+  std::string if_label = PrefixLabel("if", label_num);
```
Lai-YT, line 127:
```suggestion
  std::string then_label = PrefixLabel("then", label_num);
```
Using `then_label` would be more expressive, indicating that this label marks the start of the "then" part.

src/qbe_ir_generator.cpp
```diff
@@ -124,14 +124,28 @@ void QbeIrGenerator::Visit(const IfStmtNode& if_stmt) {
...
+  if (if_stmt.or_else) {
+    output << else_label << std::endl;
+  } else {
+    output << end_label << std::endl;
+  }
+
+  output << if_label << std::endl;
+  if_stmt.then->Accept(*this);
+  if (if_stmt.or_else) {
+    // Jumps to end_label after finishing the first body.
```
Lai-YT, line 144:
```suggestion
    // Skip the "else" part after executing "then".
```

src/qbe_ir_generator.cpp
```diff
@@ -124,14 +124,28 @@ void QbeIrGenerator::Visit(const IfStmtNode& if_stmt) {
...
 
-  // Jumps to start_label if argument is non-zero. Or else, jumps to end_label.
-  output << "jnz " << PrefixSigil(predicate_num) << ", " << start_label << ", "
-         << end_label << std::endl;
-  output << start_label << std::endl;
-  if_stmt.body->Accept(*this);
+  // Jumps to if_label if argument is non-zero. Or else, jumps to the second
+  // label.
+  output << "jnz " << PrefixSigil(predicate_num) << ", " << if_label << ", ";
+  // Output else_label as the second jump label if there is a second body.
```
Lai-YT, line 134: This comment can then be removed.
```suggestion
```

src/qbe_ir_generator.cpp
```diff
@@ -124,14 +124,28 @@ void QbeIrGenerator::Visit(const IfStmtNode& if_stmt) {
...
+  std::string else_label = PrefixLabel("else", label_num);
   std::string end_label = PrefixLabel("end", label_num);
 
-  // Jumps to start_label if argument is non-zero. Or else, jumps to end_label.
-  output << "jnz " << PrefixSigil(predicate_num) << ", " << start_label << ", "
-         << end_label << std::endl;
-  output << start_label << std::endl;
-  if_stmt.body->Accept(*this);
+  // Jumps to if_label if argument is non-zero. Or else, jumps to the second
+  // label.
```
Lai-YT, line 132: I would prefer a comment explaining the logic of the generated code instead of attempting to rephrase each line individually:
```suggestion
  // Jumps to "then" if the predicate is true (non-zero), else jumps to "else".
  // If no "else" part exists, jumps directly to "end".
  // In the presence of an "else" part, a second jump is needed after executing
  // "then" to skip it, as the generated code for "else" follows immediately
  // after "then".
```

Lai-YT (changes requested):
Here are some suggested changes, primarily related to the comments :pencil2:.
Please let me know if you have different opinions :grinning:.

src/qbe_ir_generator.cpp
leewei05: @Lai-YT 
I think using `then` as the first statement body member for `IfStmtNode` is a great idea because it has additional context.

```
if_stmt.then->Accept(*this);
```

However, I think we should keep `if_label` for this case from the perspective of QBE IR. It would be more straightforward looking at the following IR code. I think `@if.1` pairs nicely with `@else.1`, but `@then.1` doesn't provide enough context in the assembly code since there's only jump instruction at the IR level. What do you think?

Suggested change:
```
jnz %.5, @then.1, @else.1
@then.1
  %.6 =w copy 0
  storew %.6, %.1
  jmp @end.1
@else.1
  %.7 =w loadw %.1
  %.8 =w copy 1
  %.9 =w add %.7, %.8
  storew %.9, %.1
@end.1
```

Current:
```
jnz %.5, @if.1, @else.1
@if.1
  %.6 =w copy 0
  storew %.6, %.1
  jmp @end.1
@else.1
  %.7 =w loadw %.1
  %.8 =w copy 1
  %.9 =w add %.7, %.8
  storew %.9, %.1
@end.1
```

src/qbe_ir_generator.cpp
leewei05: Great suggestions! 👍🏼

src/qbe_ir_generator.cpp
leewei05: Having this may be confusing in the code, right? 😂 
```
std::string then_label = PrefixLabel("if", label_num);
```

Lai-YT: In my humble opinion, in an _if-then-else_ statement, _if_ refers to the condition, while _then_ and _else_ refer to the block statements.
That's why I propose labels like `@then.1` and `@else.1` to maintain clarity.
However, I understand your concern. How about choosing one of the following options:
- Add a comment `# if` above the evaluation of the predicate.
- Change the label of the "then" part to `@if_then.1`.

leewei05: I like option 1 better because we can increase readability in both source code and IR.
So, I will use `then_label` in the source code and add a comment above the predicate in the IR. 👍🏼

Lai-YT (approved):
Looking forward to the support for if statements without curly braces!

PR #53, opened 2023-11-30
Move block statement to statement

leewei05:
- Move block statement to statement
- Support single statement for if statement
- Fix dangling else problem


parser.y
```diff
@@ -57,15 +57,30 @@ extern std::unique_ptr<AstNode> program;
...
-%nterm <std::unique_ptr<BlockStmtNode>> main_func
+%nterm <std::unique_ptr<StmtNode>> main_func
 
 %left '='
 %left EQ NE
 %left '<' '>' LE GE
 %left '+' '-'
 %left '*' '/' '%'
 
+// Example: IF '(' expr ')' IF '(' expr ')' stmt • ELSE stmt
```
Lai-YT, line 68: Upon initial inspection, it is unclear what this example is attempting to demonstrate.
I recommend adding the following imperative sentence for clarity:
```suggestion
// Resolve the ambiguity in the "dangling-else" grammar.
// Example: IF '(' expr ')' IF '(' expr ')' stmt • ELSE stmt
```

Lai-YT (changes requested):
I believe it's essential to distinguish between blocks and normal statements since blocks cannot always be replaced by a statement. Consider the `main_func_` as an example, which must have its body enclosed within a pair of curly braces.

Allowing any statement to represent its body could result in syntactically valid, yet unintended code like the following Python-ish syntax:

```c
int main()
    return 0;
```
 
For reference, in the C standard ISO/IEC 9899:1999, section 6.9.1 defines the syntax of a _function-definition_:
- _function-definition_:
    _declaration-specifiers_ _declarator_ _declaration-list<sub>opt<sub>_ _compound-statement_

Here, the _compound-statement_ corresponds to our _block-statement_.

Lai-YT (changes requested):

PR #54, opened 2023-11-30
Rename precedence to resolve confusion

Lai-YT:
The precedence had the same name as the token, while the former one is used as a placeholder for the precedence. Therefore, it's better for them to use different names to avoid possible confusion.
The `%prec IF_WITHOUT_ELSE` simply instructs Bison that the rule `IF '(' expr ')' stmt` has the same precedence as `IF_WITHOUT_ELSE`.

Please let me know if you have any advice or suggestions on the naming! :smile:


leewei05 (approved):
LGTM! 😄

PR #55, opened 2023-12-03
Remove test target from all

leewei05:
Since we have started to have more test cases, each build will take longer with default test target.
Thus, it would be better to remove test from default targets.


Lai-YT (approved):
Agree!

PR #56, opened 2023-12-03
Implement for statement ast_dumper

leewei05:
- Parse for statement
- Implement for statement ast_dumper
- Bump build version to C++17

Please note that even if `step` is evaluated after each statement execution, we still dump and typecheck `step` before `loop_body`. It will only affect the order when generating assembly code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 TARGET := vitaminc
 CXX := g++
 CC = $(CXX)
-CXXFLAGS = -g3 -std=c++14 -Wall -MMD -Iinclude -Werror
+CXXFLAGS = -g3 -std=c++17 -Wall -MMD -Iinclude -Werror
 CFLAGS = $(CXXFLAGS)
 LEX = lex
 # C++ features are used, yacc doesn't suffice


src/ast_dumper.cpp
```diff
@@ -37,7 +37,15 @@ void AstDumper::Visit(const DeclNode& decl) {
...
-  std::cout << ')' << std::endl;
+  std::cout << indenter_.Indent() << ')' << std::endl;
+}
+
+void AstDumper::Visit(const LoopInitNode& loop_init) {
+  if (std::holds_alternative<std::unique_ptr<DeclNode>>(loop_init.clause)) {
+    std::get<std::unique_ptr<DeclNode>>(loop_init.clause)->Accept(*this);
+  } else {
+    std::get<std::unique_ptr<ExprNode>>(loop_init.clause)->Accept(*this);
+  }
```
Lai-YT, line 48: Just to let you know, the following alternative achieves the same semantics:
```cpp
  std::visit([this](const auto& node) { node->Accept(*this); },
             loop_init.clause);
```
It appears to sacrifice readability for me, so I wouldn't recommend it.

Lai-YT (commented):
Is there a test case of the _for statement_ with a _declaration_ clause?

leewei05:
> Is there a test case of the _for statement_ with a _expression_ clause?

I added one.

Lai-YT (approved):

PR #57, opened 2023-12-03
Enable parallel build and test

leewei05:
- Utilize multiple processors to speedup build.
- Enable parallel testing. In order to avoid data race in code generation test, we change the output filename.

It isn't that obvious with Github Actions, but the speed improves significantly in local build and testing. 😄


leewei05:
Still fixing warning messages in Linux.
```
warning: jobserver unavailable: using -j1. Add `+' to parent make rule.
```

PR #58, opened 2023-12-04
Enable parallel test option

leewei05:
Whenever users pass the `-j` flag with make, this setting will also enable parallel testing with turnt. For example:

```
make test -j8
```

Also, we don't want to pass `-j` to our sub-make since we only need `turnt` to execute in parallel.


Lai-YT (approved):
LGTM 🔥 
Thanks for improving our build system!

PR #59, opened 2023-12-08
Merge expression statement

leewei05:
Since we have an `expr_opt` non terminal node, we can merge `NullStmtNode` and `ExprStmtNode` into one.

```
6.8.3 Expression and null statements
Syntax
expression-statement:
  expressionopt ;
```


Lai-YT (approved):
Absolutely, I share the same perspective on this! Let's go ahead with the merge.

PR #64, opened 2023-12-14
Parse jump statements

leewei05:
- Parsing break and continue statements
- Add dump test cases

Based on C standards:

> A continue statement shall appear only in or as a loop body.
> A break statement shall appear only in or as a switch body or loop body.

For type check and code generation, these two statements will need additional context to check and produce the correct jump. We can discuss the details offline. 🖖🏼


test/typecheck/continue_stmt.c
```diff
@@ -0,0 +1,13 @@
...
+  do {
+    if (a % 2 == 0) {
+      ++a;
+      continue;
+    }
+    ++a;
+
+  } while (a < 5);
+
+  return 0;
```
Lai-YT, line 12:
```suggestion
  int a;
  do {
    if (true) {
      a = 5;
      continue;
    }
    a = 0;

  } while (a < 5);

  return 0;
```
I suggest focusing the test on `continue` itself, utilizing a simple `true` or `false` predicate with fewer kinds of operators. This approach aims to prevent false failures due to the breakage of other features. Additionally, since we are testing the syntax rather than the semantics, it's acceptable to keep the program simple.

Lai-YT (changes requested):
Is the absence of the goto statement intentional? It's one of the jump statements. :thinking:
For discussions on type checking and code generation, I'll connect with you later. :pray:

test/typecheck/continue_stmt.c
leewei05: Sure. Changed the predicate to `1` since we don't have `true`, `false` keywords yet.

Lai-YT: Oh, yes, `1`. :laughing:

Lai-YT (approved):

PR #65, opened 2023-12-15
Refactor code generation with fmt

leewei05:
- Utilize fmt to generate *.ssa file with indent.
- Link fmt library

The following is the generated output:

```
# for_stmt.ssa
export function w $main() {
@start
        %.1 =l alloc4 4
        %.2 =w copy 0
        storew %.2, %.1
        %.3 =l alloc4 4
        # loop init
        %.4 =w copy 0
        storew %.4, %.3
@pred.1
        %.5 =w loadw %.3
        %.6 =w copy 5
        %.7 =w csltw %.5, %.6
        jnz %.7, @loop_body.1, @end.1
@loop_body.1
        %.8 =w loadw %.1
        %.9 =w copy 1
        %.10 =w add %.8, %.9
        storew %.10, %.1
        %.11 =w loadw %.3
        %.12 =w copy 1
        %.13 =w add %.11, %.12
        storew %.13, %.3
        jmp @pred.1
@end.1
        %.14 =w copy 0
        ret %.14
}
```

Looks better than before! 😄

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,8 @@ test: $(TARGET)
 	cd test/ && $(MAKE) test MAKEFLAGS=
 
 $(TARGET): $(OBJS)
-	$(CXX) $(CXXFLAGS) $(OBJS) -o $@
+	# Link fmt library after all object files, or else it will throw errors.
+	$(CXX) $(CXXFLAGS) $(OBJS) -lfmt -o $@
 
 lex.yy.cpp: lexer.l
 	$(LEX) -o $@ $<
--- a/compile_flags.txt
+++ b/compile_flags.txt
@@ -1,0 +1,2 @@
+-I./include
+-I/usr/local/include
--- a/src/qbe_ir_generator.cpp
+++ b/src/qbe_ir_generator.cpp
@@ -16,8 +16,6 @@ extern std::ofstream output;
 
 namespace {
 
-const int kIndentSize = 8;
-
 /// @brief Returns the next local number and increment it by 1. The first number
 /// will be 1.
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -32,7 +32,8 @@ test: $(TARGET)
 	cd test/ && $(MAKE) test MAKEFLAGS=
 
 $(TARGET): $(OBJS)
-	$(CXX) $(CXXFLAGS) $(OBJS) -o $@
+	# Link fmt library after all object files, or else it will throw errors.
+	$(CXX) $(CXXFLAGS) $(OBJS) -lfmt -o $@
```
Lai-YT, line 36: I believe you're referring to the order problem of linking to a static library[^1].
The proper way to resolve this is to add `LDLIBS` as `-lfmt` and append it to the end of the recipe:

```suggestion
	# Link fmt library after all object files, or else it will throw errors.
	$(CXX) $(CXXFLAGS) $(OBJS) -o $@ $(LDLIBS) 
```
Libraries must be linked after specifying the object files.

[^1]: Reference readings: [c++ undefined references with static library](https://stackoverflow.com/questions/2624238/c-undefined-references-with-static-library), [Why does the order in which libraries are linked sometimes cause errors in GCC?](https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc)

Lai-YT (changes requested):

leewei05:
Handover to @Lai-YT 🙏🏼

PR #67, opened 2024-01-20
Resolve include dependency for lex.yy.cpp

Lai-YT:
Previously, we encountered issues due to the lack of forward declaration for the lexing function `yylex`. Consequently, we resorted to including lex.yy.cpp directly in parser.y to access it. The initial problem stemmed from our inability to insert it into the correct position within yy.tab.hpp. Recently, I found that the `%code` directive is aptly suited for addressing this issue.
The current dependency is now one-way, with only lex.yy.cpp depending on y.tab.hpp. Modifications have been made to the Makefile to align with this adjustment.

For more information, refer to:
https://www.gnu.org/software/bison/manual/html_node/_0025code-Summary.html#index-_0025code-5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,7 @@ endif
 # Export variable to be visible for test/Makefile.
 export PARALLEL
 
-# Note that lex.yy.cpp is excluded deliberately, as "lex.yy.cpp" is considered a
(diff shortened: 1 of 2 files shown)


leewei05 (approved):
LGTM! 👍🏼

Issue #69, opened 2024-02-21
Refactor OpGetter to Enum class


leewei05:
Currently, binary and unary operations of different operators are designed as concrete types. `BinaryExprNode` and `UnaryExprNode` act as abstract classes. However, since these operations usually share common behaviors, heavy use of polymorphism doesn't seem useful, and in fact, it can lead to boilerplate code.
We propose removing these classes, denoting different operations with an `enum`, and making `BinaryExprNode` and `UnaryExprNode` concrete classes that hold such an `enum`. Although we'll have to use `switch`-`case` statements to dispatch with them, we believe this change will provide better readability and maintainability.

PR #77, opened 2024-02-24
Refactor OpGetter with enum classes

leewei05:
1. Introduce enum class for operators and remove redundant functions.
2. I added a temporary fix for unused function error until we implement unary expression code generation. Any better suggestions?

Resolves: #69 

---

No more dispatch macros. 😸


include/ast.hpp
```diff
@@ -210,160 +211,30 @@ struct IntConstExprNode : public ExprNode {
...
-
-/// @brief This is a bitwise complement unary expression node.
-struct BitCompExprNode : public UnaryExprNode {
-  using UnaryExprNode::UnaryExprNode;
-
-  virtual void Accept(NonModifyingVisitor&) const override;
-  virtual void Accept(ModifyingVisitor&) override;
-};
-
 /// @note This is an abstract class.
```
Lai-YT, line 224: `BinaryExprNode` is now a concrete class.

include/ast.hpp
```diff
@@ -210,160 +211,30 @@ struct IntConstExprNode : public ExprNode {
 
 /// @note This is an abstract class.
```
Lai-YT, line 212: `UnaryExprNode` is now a concrete class.

src/qbe_ir_generator.cpp
```diff
@@ -37,6 +38,50 @@ void VWriteOut(fmt::string_view format, fmt::format_args args) {
...
+    case BinaryOperator::kLt:
+      return "csltw";
+    case BinaryOperator::kLte:
+      return "cslew";
+    case BinaryOperator::kEq:
+      return "ceqw";
+    case BinaryOperator::kNeq:
+      return "cnew";
+    default:
+      return "Unknown";
```
Lai-YT, line 67: We are returning an "unknown" type that does not start with a capital letter for our types:
https://github.com/fruits-lab/VitaminC/blob/4d6a6960ce60c032406b6f5e08d8616b3759ac4d/src/type.cpp#L5-L13

Do you think it's better if we maintain consistency in this regard? Alternatively, should we throw an error (or trigger an assertion failure) here, since `ExprType::kUnknown` is valid, but operators are not?

include/ast.hpp
leewei05: Nice catch!

src/qbe_ir_generator.cpp
leewei05: As we discussed, we will change unknown operator to unreachable in the next PR.

Lai-YT (approved):
LGTM :rocket:
Thanks for the revision!

PR #81, opened 2024-02-25
Mix declarations and statements in compound statement

Lai-YT:
This enhancement allows for declarations and statements to interleave within compound statements.
I follow the reference manual, naming the "nodes" inside the compound statement as `item`.

> [!note]
> Previously, there was a discussion regarding the use of `std::visit` when dispatching on `std::variant`. Initially, `std::visit` was considered less readable (see https://github.com/fruits-lab/VitaminC/pull/56#discussion_r1418991588).
> Upon reconsideration, it was realized that the actual underlying type is irrelevant, as the types are polymorphic and always call the `Accept` function. Therefore, using `std::visit` hides unnecessary detail, which is preferable to verbose holding conditions using `std::holds_alternative`.

Resolves: #79

### DIFF
--- a/include/ast.hpp
+++ b/include/ast.hpp
@@ -88,15 +88,14 @@ struct LoopInitNode : public AstNode {
 };
 
 struct CompoundStmtNode : public StmtNode {
-  CompoundStmtNode(std::vector<std::unique_ptr<DeclNode>>&& decls,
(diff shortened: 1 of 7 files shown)


leewei05 (approved):
LGTM! 😄

PR #82, opened 2024-02-25
Exclude test files from tidy check

Lai-YT:
Previously, clang-tidy-diff was applied to all changed files, including those under the test/ directory. However, these files are not production code and may contain unusual constructs. Therefore, we exclude them from the output of `git diff`, ensuring they are not checked by clang-tidy.

For more details, please refer to [Making 'git log' ignore changes for certain paths](https://stackoverflow.com/questions/5685007/making-git-log-ignore-changes-for-certain-paths/21079437#21079437).


leewei05 (approved):
LGTM!

Lai-YT:
> LGTM!

Thanks for reviewing!

PR #83, opened 2024-02-25
Fix missing exclude mark in pathspec

Lai-YT:
Unfortunately, I made a mistake and failed to properly test it in #82.


leewei05 (approved):
Gogo!

PR #84, opened 2024-02-25
Drop build of examples/tests of library to speed up installation

Lai-YT:
By default, the tests of fmt and the examples/tests of cxxopts are also built during installation, which slows down our CI process. Since we will not run the fmt tests or those of cxxopts, building them is unnecessary.


leewei05 (approved):
Gogo!

leewei05:
Nice fix! 😄

Issue #85, opened 2024-02-27
Support function feature


leewei05:
- [x] Function calls (postfix-expressions)
- [x] Function definition
- [x] Parameters lists
- [x] Parameter lists code generation
- [x] Function definition code generation
- [x] Function call code generation
- [x] Function call with arguments code generation
- [x] Function arguments
- [ ] Type check for arguments and parameters

PR #87, opened 2024-02-27
Parse function definition

leewei05:
This PR only parse a function with no parameters. I will introduce function call and support parameters in future PRs.

Worth noting that this change create a new symbol table before parsing functions (storing function ids).

Related to #85

### DIFF
--- a/include/ast.hpp
+++ b/include/ast.hpp
@@ -75,6 +75,19 @@ struct DeclNode : public AstNode {
   std::unique_ptr<ExprNode> init;
 };
 
+struct FuncDefNode : public AstNode {
+  FuncDefNode(std::string id, std::unique_ptr<CompoundStmtNode> body,
+              ExprType return_type)
+      : id{std::move(id)}, body{std::move(body)}, return_type{return_type} {}
+
+  void Accept(NonModifyingVisitor&) const override;
+  void Accept(ModifyingVisitor&) override;
+
+  std::string id;
+  std::unique_ptr<CompoundStmtNode> body;
+  ExprType return_type;
+};
+
 /// @brief A loop initialization can be either a declaration or an expression.
 struct LoopInitNode : public AstNode {
   LoopInitNode(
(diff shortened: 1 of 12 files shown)


parser.y
```diff
@@ -102,8 +104,22 @@
 %start entry
 
 %%
-entry: main_func {
-    program = std::make_unique<ProgramNode>($1);
+// TODO: support global variables
+entry: func_decl_list_opt main_func {
```
Lai-YT, line 108: Can we merge the `main_func` into the `func_decl_list_opt` at this stage?

Lai-YT (commented):
I'm a bit confused about whether this PR aims to parse a function declaration or a function definition. Here's the syntax of a `translation-unit`, which references our `entry`:

translation-unit:
- external-declaration
- translation-unit external-declaration

external-declaration:
- function-definition
- declaration

I suspect this PR is actually focusing on the function definition. Am I correct?

parser.y
Lai-YT: The merge will be deferred to a later PR. :love_you_gesture:

include/ast.hpp
```diff
@@ -75,6 +75,19 @@ struct DeclNode : public AstNode {
   std::unique_ptr<ExprNode> init;
 };
 
+struct FuncDeclNode : public AstNode {
```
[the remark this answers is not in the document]
leewei05: Fixed!

parser.y
Lai-YT, line 108: Update as `func_def_list_opt`.

src/qbe_ir_generator.cpp
```diff
@@ -123,6 +123,8 @@ void QbeIrGenerator::Visit(const DeclNode& decl) {
   id_to_num[decl.id] = id_num;
 }
 
+void QbeIrGenerator::Visit(const FuncDefNode& decl) {}
```
Lai-YT, line 126:
```suggestion
void QbeIrGenerator::Visit(const FuncDefNode& def) {}
```

src/type_checker.cpp
```diff
@@ -23,6 +23,19 @@ void TypeChecker::Visit(DeclNode& decl) {
   }
 }
 
+void TypeChecker::Visit(FuncDefNode& func_decl) {
```
Lai-YT, line 26:
```suggestion
void TypeChecker::Visit(FuncDefNode& func_def) {
```

include/ast.hpp
```diff
@@ -101,11 +114,14 @@ struct CompoundStmtNode : public StmtNode {
 /// @brief Root of the entire program.
 struct ProgramNode : public AstNode {
   /// @note vector of move-only elements are move-only
-  ProgramNode(std::unique_ptr<CompoundStmtNode> body) : body{std::move(body)} {}
+  ProgramNode(std::vector<std::unique_ptr<FuncDefNode>> func_decl_list,
```
Lai-YT, line 117:
```suggestion
  ProgramNode(std::vector<std::unique_ptr<FuncDefNode>> func_def_list,
```

Lai-YT (changes requested):
A few renames are necessary due to the change in the name of the AST node.
You can use `grep func_decl -rn ./` to check if there are any occurrences that were missed. :sunglasses:

Lai-YT (approved):
LGTM!

PR #89, opened 2024-02-28
Add bison flags to generate better error messages

leewei05:
`-Wcounterexample` generates the following example for solving conflicts.

```
Example: INCR expr • '+' expr
  Shift derivation
    expr
    ↳ 29: unary_expr
          ↳ 43: INCR expr
                     ↳ 30: expr • '+' expr
  Reduce derivation
    expr
    ↳ 30: expr                    '+' expr
          ↳ 29: unary_expr
                ↳ 43: INCR expr •
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ LEX = lex
 # C++ features are used, yacc doesn't suffice
 YACC = bison
 # -d: generate header with default name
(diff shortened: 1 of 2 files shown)


Lai-YT (approved):
Nice fix!

PR #91, opened 2024-02-28
Generate code of break statement

Lai-YT:
There are three types of loop constructs in C, all of which are tested in both nested (2 levels) and non-nested forms.

Resolves: #90


src/qbe_ir_generator.cpp
```diff
@@ -237,7 +258,10 @@ void QbeIrGenerator::Visit(const ReturnStmtNode& ret_stmt) {
   WriteOut_("ret {}\n", FuncScopeTemp{ret_num});
 }
 
-void QbeIrGenerator::Visit(const BreakStmtNode& break_stmt) {}
+void QbeIrGenerator::Visit(const BreakStmtNode& break_stmt) {
+  assert(!label_views_of_jumpable_blocks.empty());
```
leewei05, line 262: This should already be checked by TypeChecker whether this break statement is in a loop, am I right? Just to clarify, it's a runtime check for this list?

leewei05 (changes requested):

src/qbe_ir_generator.cpp
leewei05: Ha, seems like you implemented it in #92. 😂

Lai-YT: > This should already be checked by TypeChecker whether this break statement is in a loop, am I right? Just to clarify, it's a runtime check for this list?

Yes, you're correct. The `assert` statement here serves to enforce that the condition should always hold true. If it's violated, it indicates an internal compiler error.

parser.y
```diff
@@ -161,15 +161,21 @@ decl: INT ID ';' { $$ = std::make_unique<DeclNode>($2, ExprType::kInt); }
...
     | IF '(' expr ')' stmt ELSE stmt { $$ = std::make_unique<IfStmtNode>($3, $5, $7); }
     | WHILE '(' expr ')' stmt { $$ = std::make_unique<WhileStmtNode>($3, $5); }
     | DO stmt WHILE '(' expr ')' ';' { $$ = std::make_unique<WhileStmtNode>($5, $2, true); }
     | FOR '(' loop_init expr_opt ';' expr_opt ')' stmt { $$ = std::make_unique<ForStmtNode>($3, $4, $6, $8); }
+    | jump_stmt { $$ = $1; }
+    ;
+
+/* 6.8.6 Jump statements */
+jump_stmt:
+    RETURN expr ';' { $$ = std::make_unique<ReturnStmtNode>($2); }
```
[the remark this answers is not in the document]
Lai-YT: Sure. We can be consistent on this.

leewei05 (approved):
LGTM! 🚀

PR #92, opened 2024-02-28
Check whether break statement is in body of loop

Lai-YT:
# What changed?

A data structure has been introduced to the type checker to track the type of body each node is currently in.

# How to test?

Due to limitations in our test infrastructure, no test cases have been added to the test directory. However, you can manually verify the functionality by compiling the following C program with _VitaminC_:

```c
int main() {
  break;
  return 0;
}
```

If our compiler effectively captures the error, you should encounter an assertion failure `virtual void TypeChecker::Visit(BreakStmtNode&)`.

> [!note]
> Subsequent pull requests that involve constructs entering a body (compound statement) should update this data structure with their respective body type.

### DIFF
--- a/src/type_checker.cpp
+++ b/src/type_checker.cpp
@@ -1,9 +1,35 @@
 #include "type_checker.hpp"
 
+#include <algorithm>
+#include <cassert>
+#include <cstdint>
 #include <memory>
 #include <variant>
+#include <vector>
 
 #include "ast.hpp"
+
+namespace {
+
+/// @brief Some statements can only appear in body of certain constructs, namely
+/// the return, break, and continue statements.
+enum class BodyType : std::uint8_t {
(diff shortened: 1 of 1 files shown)


src/type_checker.cpp
```diff
@@ -61,22 +87,28 @@ void TypeChecker::Visit(ProgramNode& program) {
 
 void TypeChecker::Visit(IfStmtNode& if_stmt) {
   if_stmt.predicate->Accept(*this);
+  body_types.push_back(BodyType::kLoop);
```
leewei05, line 90: Why is `If` statement a `kLoop`, but not a `kNone`?

Lai-YT: Oops. :cry: You're right. The `If` statement should not have a body type. I'll address it.

PR #97, opened 2024-03-02
Generate code of continue statement

Lai-YT:
As the type checking of the continue statement is straightforward, it is integrated into this pull request.

Additionally, for the `ForStmtNode`, a `step` label has been introduced. This ensures that when a `continue` statement is encountered, the loop variable can still be advanced correctly.

> [!note]
> _cproc_ also generates a `step` label.

Resolves: #96


leewei05 (approved):
LGTM!

Issue #99, opened 2024-03-02
Indent generated QBE IR


Lai-YT:
Mixing labels with instructions can make the IR harder to read. We propose indenting the instructions so that they are grouped like paragraphs, which should improve readability. 😄

PR #100, opened 2024-03-02
Indent generated QBE IR

Lai-YT:
Two functions have been added to implement the indentation logic for labels and instructions, respectively. The pure `Write` function is retained for manual formatting purposes.

Every instruction should be indented by one level; label definitions and comments should not be indented.

> [!important]
> For future QBE IR generation, utilize `WriteLabel_` to write label definitions and `WriteInstr_` to write instructions. Both functions generate a single line, eliminating the need to manually add newline characters (`\n`) to your `format` calls. When writing parts of an instruction or multiple instructions, use `Write_` and remember to prepend each instruction with `kIndentStr` and enclose them with newlines.

Resolves: #99

### DIFF
--- a/include/qbe_ir_generator.hpp
+++ b/include/qbe_ir_generator.hpp
@@ -4,8 +4,10 @@
 #include <fmt/core.h>
 
 #include <iosfwd>
+#include <utility>
 
 #include "ast.hpp"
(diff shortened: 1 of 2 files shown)


leewei05 (approved):
LGTM! 🚀

PR #107, opened 2024-03-03
Introduce type specifier

leewei05:
Before implementing parameter list, adding type specifier may reduce future changes for `INT`.


Lai-YT (approved):
Nice!

Issue #106, opened 2024-03-03
Support switch feature


Lai-YT:
- [x] Type checking (#108)
- [x] Code generation (#112)

PR #108, opened 2024-03-03
Check type of switch statement

Lai-YT:
> [!note]
> - The `LabeledStmtNode` should be converted into a concrete node once support for the `goto` statement is implemented.
> - While the `case` statement should be followed by a _constant expression,_ we currently accept a general _expression_ to reduce the complexity of the grammar. This issue should be addressed in subsequent PRs.

Related: #106


src/type_checker.cpp
```diff
@@ -140,6 +140,51 @@ void TypeChecker::Visit(ContinueStmtNode& continue_stmt) {
...
+    // TODO: expression is not an integer constant expression
+  }
+  case_stmt.stmt->Accept(*this);
+}
+
+void TypeChecker::Visit(DefaultStmtNode& default_stmt) {
+  if (!IsInBodyOf(BodyType::kSwitch)) {
+    // TODO: 'default' statement not in switch statement
+  }
+  assert(!switch_already_has_default.empty());
```
leewei05, line 181: I think the logic here is a bit confusing because it feels unintuitive at first. I would think that `switch_already_has_default` should be empty. Then I checked line 158, realizing that it is correct.

src/type_checker.cpp
```diff
@@ -140,6 +140,51 @@ void TypeChecker::Visit(ContinueStmtNode& continue_stmt) {
...
+  if (switch_already_has_default.back()) {
```
leewei05, line 182: Wouldn't `default_stmt` need to push a new `true` node into `switch_already_has_default`?

src/type_checker.cpp
```diff
@@ -140,6 +140,51 @@ void TypeChecker::Visit(ContinueStmtNode& continue_stmt) {
   }
 }
 
+namespace {
+/// @brief A shared state to convey the presence of a default label in a switch
+/// statement.
+/// @note To allow nested switch statements, the state is stacked.
+auto
+    switch_already_has_default  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
```
leewei05, line 148: I think `switch_has_default` should be clear enough.

src/type_checker.cpp
leewei05: I cannot think of a better solution in the moment, maybe we just leave it.

Lai-YT: Maybe we add a comment to clarify that this sharing state is only meaningful for `switch`es?
Since the AST tree is quite general, I couldn't find a way to not let other nodes access the `switch_already_has_default`.

src/type_checker.cpp
Lai-YT: I believe that the word `already` better conveys that the default shouldn't appear twice.
Consider the use case:

```c++
if (switch_has_default.back()) {
  // error
}
```

and

```c++
if (switch_already_has_default.back()) {
  // error
}
```

src/type_checker.cpp
Lai-YT: Sorry for the confusion. The current state should be set to `true` instead of pushing a new one.

leewei05 (approved):

src/type_checker.cpp
Lai-YT: Also our testing infrastructure doesn't support testing on failed cases. :sob:
I'll skip adding the test case.

PR #110, opened 2024-03-06
Move parameter generations to its visit function

Lai-YT:
Handling the type and format of parameters should be done in the visit function of `ParamNode` rather than within `FuncDefNode`. This approach groups the logic more clearly and provides a clear indication of where to find related code.
Ideally, code generation related to a specific AST node should be performed within its visit function. Only in rare cases where information from the top level is required should generation be handled at a higher level.

Additionally, function definitions, whether with or without parameters, can be handled uniformly. The for loop simply does nothing if the parameter list is empty.


leewei05 (approved):
LGTM!

PR #111, opened 2024-03-07
Implement function call with arguments

leewei05:
Related to #85


src/type_checker.cpp
Lai-YT, line 1: Are we checking the types between the arguments and the parameters in this PR? Or would you like to add it as a TODO?

leewei05: I'll leave a TODO in this PR. 🖖🏼

Lai-YT (approved):
LGTM :hugs:

PR #112, opened 2024-03-09
Generate code of switch statement

Lai-YT:
The structure of a switch statement, including the labeled statements inside, is represented in the following pseudo IR:

```llvm
%1 = ... case ctrl ...
jmp @cond.1
; { The enclosing curly brace of the switch statement.
 ... unreachable code ...
@case.2:
 ... body ...
 jmp @exit  ; break
@case.3:
 ... body ...
@default.4:
 ... body ...
@case.5:
 ... body ...
 jmp @exit  ; break
; } The enclosing curly brace of the switch statement.
; The jmp is intentionally placed, so that the conditions are never touch twice.
; A label to avoid invalid consecutive jumps.
@switch_bottom.6
 jmp @exit
@cond.1:
 %2 =w ... case expr ...
 %3 =w ceqw %1, %2
  jnz %3, @case.2, @cond.7
@cond.7:
 %4 =w ... case expr ...
 %5 =w ceqw %1, %4
 jnz %5, @case.3, @cond.8
@cond.8:
 %8 =w ... case expr ...
 %9 =w ceqw %1, %8
; If no case matches, jump to the default case.
; If no default case exists, jump to the exit label instead.
 jnz %9, @case.5, @default.4
@exit:
 ... statements after the switch statement ...
```

The first part contains the case labels and the default label, while the second part contains the conditions matching the control expression with the case expressions.
The switch statement first jumps to the second part, which then jumps back to the first part. This ensures that each part is executed only once.

- Case labels are generated first to collect information needed for the conditions.
- Evaluation of case expressions is done in the condition part.

> [!note]
> Two consecutive jumps appear to be invalid for QBE IR (`label or } expected`), indicating that we cannot handle two consecutive break statements simply by generating two jumps. This issue may need to be addressed in future PRs.

Related: #106


leewei05 (approved):
LGTM! Great work! 🚀 🪨

PR #114, opened 2024-03-10
Check function arguments and parameters types

leewei05:
Resolves: #85

### DIFF
--- a/src/type_checker.cpp
+++ b/src/type_checker.cpp
@@ -227,7 +227,7 @@ void TypeChecker::Visit(ArgExprNode& arg_expr) {
   arg_expr.type = arg_expr.arg->type;
 }
 
-void TypeChecker::Visit(FunCallExprNode& call_expr) {
+void TypeChecker::Visit(FuncCallExprNode& call_expr) {
   call_expr.func_expr->Accept(*this);
   call_expr.type = call_expr.func_expr->type;
 
--- a/src/type_checker.cpp
+++ b/src/type_checker.cpp
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cstddef>
 #include <cstdint>
 #include <memory>
 #include <utility>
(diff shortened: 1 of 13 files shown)


src/type_checker.cpp
```diff
@@ -67,15 +68,18 @@ void TypeChecker::Visit(FuncDefNode& func_def) {
...
   } else {
+    std::vector<ExprType> param_types{};
+    for (auto& parameter : func_def.parameters) {
+      parameter->Accept(*this);
+      param_types.push_back(parameter->type);
+    }
+
     auto symbol = std::make_unique<SymbolEntry>(func_def.id);
     symbol->expr_type = func_def.return_type;
+    symbol->param_types = param_types;
```
Lai-YT, line 79: Is there any particular reason preventing you from constructing the symbol beforehand and directly pushing the parameter types back into `symbol->param_types` during the visitation of parameters?

Here's the code that expresses my intention:

```cpp
    auto symbol = std::make_unique<SymbolEntry>(func_def.id);
    symbol->expr_type = func_def.return_type;
    for (auto& parameter : func_def.parameters) {
      parameter->Accept(*this);
      symbol->param_types.push_back(parameter->type);
    }
    env_.Add(std::move(symbol));
```

I understand that these approaches differ slightly in error handling. Are we considering this aspect?

src/type_checker.cpp
```diff
@@ -229,9 +233,20 @@ void TypeChecker::Visit(FunCallExprNode& call_expr) {
...
+  auto func_def = env_.LookUp(id_expr->id);
+  auto& param_types = func_def->param_types;
+  auto& args = call_expr.args;
+  if (param_types.size() != args.size()) {
+    // TODO: argument size doesn't match
+  }
+
+  for (auto i = std::size_t{0}; i < args.size(); ++i) {
+    args[i]->Accept(*this);
+    if (args[i]->type != param_types[i]) {
```
Lai-YT, line 247:
```suggestion
    args.at(i)->Accept(*this);
    if (args.at(i)->type != param_types.at(i)) {
```

Lai-YT (changes requested):

src/type_checker.cpp
leewei05: This one is better!

Lai-YT (approved):
LGTM :rocket:

PR #117, opened 2024-03-11
Fix Turnt not running in parallel when `-j` option is passed

Lai-YT:
Originally, our test tool, _Turnt_, didn't run in parallel even when the `-j` option was passed as a command line argument to `make test -j`. This was because the `MAKEFLAGS` variable never contained the `-j` option when checked. Based on my experimentation, it seems that the `-j` option is only added to the `MAKEFLAGS` after entering the recipe if it exists. Therefore, we need to check whether the `-j` option is passed inside the recipe. This also prevents the check from polluting the Makefile, as the variable `PARALLEL` is only used by the `test` target.

Additionally, I moved the checking logic down to the Makefile under `test/`. It is more suitable for such a Makefile to know how _Turnt_ works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,13 +10,6 @@ LEX = lex
 YACC = bison
 # -d: generate header with default name
 YFLAGS = --verbose --debug -d -Wcounterexamples -Wall
-# Check if -j multiple jobs is configured.
-PARALLEL = false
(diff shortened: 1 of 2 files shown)


leewei05 (approved):
🚀

PR #118, opened 2024-03-13
Update codegen tests with builtin print

Lai-YT:
> [!note]
> The reviewer may take a quick overview of the test cases to see how the print function is utilized. Further test cases may also follow this approach.

Removed two types of tests:
1. Single statement tests: Several statements accept a statement as their inner construct, which can be either a compound statement or a single statement. Testing both cases for each statement is unnecessary; as long as the compound statement is implemented correctly, it should instruct its inner statement to perform correctly. Therefore, testing with either should have the same effect.
2. Integer constant tests: These are considered too basic and are already involved in various tests as building blocks.

See also: #115


leewei05 (approved):
LGTM! 🚀

Issue #116, opened 2024-03-11
Support pointer feature


leewei05:
- [x] parse pointer
- [x] code gen pointer
- [x] code gen pointer arguments (#135)
- [x] support pointer to pointer (#138)
- [x] support function pointer (#147)

Lai-YT:
Trying to work on arguments... 🤓

PR #119, opened 2024-03-16
Support basic pointer features

leewei05:
This PR includes the following changes:

- Pointer variables declaration
- Function parameters, arguments as pointers
- Check pointer compatibility
- Create Type struct for both `Declnode` and `ExprNode`
- Update default Type for label identifier

Function pointer and pointer to pointer type will be added in future PRs. Also, Function return pointer type may require the feature of `malloc` and `free`, which are not supported yet.

See: #116

### DIFF
--- a/src/type_checker.cpp
+++ b/src/type_checker.cpp
@@ -40,10 +40,6 @@ bool IsInBodyOf(BodyType type) {
 void TypeChecker::Visit(DeclNode& decl) {
   if (decl.init) {
     decl.init->Accept(*this);
-    if (decl.is_pointer != decl.init->is_pointer) {
-      // TODO: incompatible pointer types
-    }
-
     if (decl.init->type != decl.type) {
       // TODO: incompatible types when initializing type 'type' using type
       // 'init->type'
@@ -53,8 +49,7 @@ void TypeChecker::Visit(DeclNode& decl) {
   if (env_.Probe(decl.id)) {
     // TODO: redefinition of 'id'
   } else {
-    auto symbol = std::make_unique<SymbolEntry>(decl.id, decl.is_pointer);
-    symbol->expr_type = decl.type;
+    auto symbol = std::make_unique<SymbolEntry>(decl.id, decl.type);
     env_.Add(std::move(symbol));
   }
 }
@@ -63,9 +58,7 @@ void TypeChecker::Visit(ParamNode& parameter) {
   if (env_.Probe(parameter.id)) {
     // TODO: redefinition of 'id'
   } else {
-    auto symbol =
-        std::make_unique<SymbolEntry>(parameter.id, parameter.is_pointer);
-    symbol->expr_type = parameter.type;
+    auto symbol = std::make_unique<SymbolEntry>(parameter.id, parameter.type);
     env_.Add(std::move(symbol));
   }
 }
@@ -74,13 +67,10 @@ void TypeChecker::Visit(FuncDefNode& func_def) {
   if (env_.Probe(func_def.id)) {
     // TODO: redefinition of function id
   } else {
-    auto symbol = std::make_unique<SymbolEntry>(func_def.id);
-    symbol->expr_type = func_def.return_type;
+    auto symbol = std::make_unique<SymbolEntry>(func_def.id, func_def.type);
     for (auto& parameter : func_def.parameters) {
       parameter->Accept(*this);
-      auto param_type =
-          std::make_unique<ParamType>(parameter->type, parameter->is_pointer);
-      symbol->param_types.push_back(std::move(param_type));
+      symbol->param_types.push_back(parameter->type);
     }
 
     env_.Add(std::move(symbol));
@@ -107,10 +97,9 @@ void TypeChecker::InstallBuiltins_(ScopeStack& env) {
   // The supported builtins are:
   // - int __builtin_print(int)
 
-  auto symbol = std::make_unique<SymbolEntry>("__builtin_print");
-  symbol->expr_type = ExprType::kInt;
-  auto param_type = std::make_unique<ParamType>(ExprType::kInt);
-  symbol->param_types.push_back(std::move(param_type));
+  auto symbol =
+      std::make_unique<SymbolEntry>("__builtin_print", PrimitiveType::kInt);
+  symbol->param_types.emplace_back(PrimitiveType::kInt);
   env.Add(std::move(symbol));
 }
 
@@ -157,7 +146,7 @@ void TypeChecker::Visit(ForStmtNode& for_stmt) {
 
 void TypeChecker::Visit(ReturnStmtNode& ret_stmt) {
   ret_stmt.expr->Accept(*this);
-  if (ret_stmt.expr->type != ExprType::kInt) {
+  if (ret_stmt.expr->type != PrimitiveType::kInt) {
     // TODO: return value type does not match the function type
   }
 }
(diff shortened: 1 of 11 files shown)


Lai-YT (commented):
Regarding the design decision on the structure of expression type, have you considered making it an individual `struct` instead of a derived class of `AstNode`?

The type system represents semantic concepts. It seems somewhat unreasonable to have the expression type as a syntactic node construct. Moreover, there's a duplication of data members (`PrimitiveType type` and `bool is_pointer`) between `ExprNode` and `ExprTypeNode`. This duplication suggests an awareness that `ExprNode` holds semantic type, while `ExprTypeNode` represents type information found in syntax (grammar). It's unlikely that we need to perform analysis or generate code directly on `ExprTypeNode`s.

Another concern with `ExprTypeNode` being a derived class of `AstNode` is the loss of the ability to copy the type since it becomes a polymorphic object.

To summarize, I suggest making the type a `struct` held by both `DeclNode` and `ExprNode`. This allows us to pass such a struct by value, eliminating the need for `std::shared_ptr` (although we may need this again when supporting pointer-to-pointers). Below is the concept of the proposed code:

```cpp
struct Type {
    PrimitiveType prim_type;
    bool is_ptr;

    // If type equality is supported.

    bool operator==(const Type& that) {
        return prim_type == that.prim_type && is_ptr == that.is_ptr;
    }

    bool operator!=(const Type& that) {
        return !operator==(that);
    }

    Type(PrimitiveType prim_type = PrimitiveType::kError, bool is_ptr = false);
};

**Review by Lai-YT — CHANGES_REQUESTED** [review]
The following test fails:

```c
int main() {
  int a = 10;
  int* b;
  int* c = &a;
  b = c;

  return *c;
}
```
```
ProgramNode
  FuncDefNode main: int
    CompoundStmtNode
      DeclNode a: int
        IntConstExprNode 10: int
      DeclNode b: int *
      DeclNode c: int *
        UnaryExprNode int &
          IdExprNode a: int
      ExprStmtNode
        SimpleAssignmentExprNode int *
          b: int *
          IdExprNode c: int *
      ReturnStmtNode
        UnaryExprNode int *
          IdExprNode c: int *
```

The issue lies in the output of `AstDumper`. Specifically, the type of the simple assignment should be `int *`, but it's not being properly handled for pointers, as well as the type of variable `b`. We need to ensure that we always convert the primitive type of a string when dealing with pointers. :smile:

src/type_checker.cpp
```diff
@@ -99,9 +97,10 @@ void TypeChecker::InstallBuiltins_(ScopeStack& env) {
   // The supported builtins are:
   // - int __builtin_print(int)
 
-  auto symbol = std::make_unique<SymbolEntry>("__builtin_print");
-  symbol->expr_type = ExprType::kInt;
-  symbol->param_types.push_back(ExprType::kInt);
+  auto type = Type(PrimitiveType::kInt);
+  auto symbol = std::make_unique<SymbolEntry>("__builtin_print", type);
+  auto param_type = Type(PrimitiveType::kInt);
+  symbol->param_types.push_back(param_type);
```
Lai-YT, line 103:
```suggestion
  auto symbol = std::make_unique<SymbolEntry>("__builtin_print", PrimitiveType::kInt);
  symbol->param_types.emplace_back(PrimitiveType::kInt);
```

src/type_checker.cpp
```diff
@@ -148,7 +147,7 @@ void TypeChecker::Visit(ForStmtNode& for_stmt) {
 
 void TypeChecker::Visit(ReturnStmtNode& ret_stmt) {
   ret_stmt.expr->Accept(*this);
-  if (ret_stmt.expr->type != ExprType::kInt) {
+  if (ret_stmt.expr->type.prim_type != PrimitiveType::kInt) {
```
Lai-YT, line 150:
```suggestion
  if (ret_stmt.expr->type != PrimitiveType::kInt) {
```

src/type_checker.cpp
```diff
@@ -182,7 +181,7 @@ auto
 
 void TypeChecker::Visit(SwitchStmtNode& switch_stmt) {
   switch_stmt.ctrl->Accept(*this);
-  if (switch_stmt.ctrl->type != ExprType::kInt) {
+  if (switch_stmt.ctrl->type.prim_type != PrimitiveType::kInt) {
```
Lai-YT, line 184:
```suggestion
  if (switch_stmt.ctrl->type != PrimitiveType::kInt) {
```

src/type_checker.cpp
```diff
@@ -198,7 +197,8 @@ void TypeChecker::Visit(IdLabeledStmtNode& id_labeled_stmt) {
   // TODO: To allow forward reference, the label has to be added in some other
   // way.
   // TODO: The label has function scope.
-  env_.Add(std::make_unique<SymbolEntry>(id_labeled_stmt.label));
+  auto label_type = Type();
+  env_.Add(std::make_unique<SymbolEntry>(id_labeled_stmt.label, label_type));
```
Lai-YT, line 201: Inline the variable since its already clear that its the type of the label.
Addionally, having labels as error type is unusual. I suggest adding a comment.
```suggestion
  // XXX: Labels have error type.
  env_.Add(std::make_unique<SymbolEntry>(id_labeled_stmt.label, PrimitiveType::kError));
```

src/type_checker.cpp
```diff
@@ -207,7 +207,7 @@ void TypeChecker::Visit(CaseStmtNode& case_stmt) {
     // TODO: 'case' statement not in switch statement
   }
   case_stmt.expr->Accept(*this);
-  if (case_stmt.expr->type != ExprType::kInt) {
+  if (case_stmt.expr->type.prim_type != PrimitiveType::kInt) {
```
Lai-YT, line 210:
```suggestion
  if (case_stmt.expr->type != PrimitiveType::kInt) {
```

src/type_checker.cpp
```diff
@@ -242,7 +242,7 @@ void TypeChecker::Visit(IdExprNode& id_expr) {
 }
 
 void TypeChecker::Visit(IntConstExprNode& int_expr) {
-  int_expr.type = ExprType::kInt;
+  int_expr.type.prim_type = PrimitiveType::kInt;
```
Lai-YT, line 245:
```suggestion
  int_expr.type = Type{PrimitiveType::kInt};
```

src/type_checker.cpp
```diff
@@ -273,7 +273,21 @@ void TypeChecker::Visit(FuncCallExprNode& call_expr) {
 
 void TypeChecker::Visit(UnaryExprNode& unary_expr) {
   unary_expr.operand->Accept(*this);
+  // unary_expr.type.prim_type = unary_expr.operand->type.prim_type;
```
Lai-YT, line 276: Remove commented code.

Lai-YT (changes requested):
The most important fact is that our `Type` can be implicitly converted from `PrimitiveType`. This means that if we test equality between them, for example:

```cpp
if (expr.type == PrimitiveType::kInt) {}
```
From the perspective of the compiler, we are actually doing:
```cpp
if (expr.type == Type{PrimitiveType::kInt}) {}
```
Thus, it's valid. This also implies that we can construct a Type from `PrimitiveType`:
```cpp
Type type = PrimitiveType::kInt;
// Although the following is preferred:
auto type = Type{PrimitiveType::kInt};
```

Last but not least, we prefer to use brace initialization on constructors:
```
// Good
auto type = Type{};

// Bad
auto type = Type();
```
For more information on the advantages of brace initialization, see [this Stack Overflow post](https://stackoverflow.com/questions/18222926/what-are-the-advantages-of-list-initialization-using-curly-braces).

> [!note]
> We may consider adding the [explicit specifier](https://en.cppreference.com/w/cpp/language/explicit) to the constructor to disable implicit conversions, but I find it convenient and still reasonable in this case.

Lai-YT (approved):
LGTM!
Nice to see pointers getting on the way. :partying_face:

Issue #120, opened 2024-03-16
Support goto feature


Lai-YT:
- [x] Type checking (#121)
- [ ] Code generation (#122)

PR #121, opened 2024-03-17
Check type of goto statement

Lai-YT:
> [!note]
> The label has function scope, which can be referenced before being declared. As a quick workaround, in this PR, the goto statement always assumes that the label is declared (or will be declared).
Note that _cproc_ also doesn't check the existence of the label: it generates a jump to an undefined label even if the label isn't declared.

Related to #120


leewei05 (approved):
LGTM!

Issue #123, opened 2024-03-19
Separate user-defined sigils from compiler-generated sigils


Lai-YT:
# What's the Problem?

It seems that sigils are used in two different scenarios:

1. Compiler-generated sigils, like labels employed for controlling loop flow.
2. User-defined sigils, such as labels used by the `goto` statement or function names.

Although these scenarios are unlikely to conflict due to their unique local numbers, I propose a differentiation strategy: all compiler-generated sigils have a `.` before the name, while user-defined sigils do not.

```
@.compiler-generated
@user-defined
```

Currently, compiler-generated scenarios are hard-coded, which is less than optimal:

https://github.com/fruits-lab/VitaminC/blob/be04a5ab65c823afbce8203d17d98c91c2c0864f/src/qbe_ir_generator.cpp#L517-L522

# Possible Implementation

Add another layer of inheritance to `Sigil`.

PR #125, opened 2024-03-23
Separate user defined sigils from compiler generated sigils

Lai-YT:
Instead of adding another layer to the inheritance, I refactored the sigils to use template specializations due to the following reasons:
- The kind of sigil is unlikely to be extended.
- All sigils have the same interface and behavior, except for the prefix.
- Polymorphic objects cannot be easily copied or moved.

I also believe that it's simpler and clearer to understand that we have four kinds of sigils:

```cpp
using BlockLabel = Sigil<'@'>;
using GlobalPointer = Sigil<'$'>;
using FuncScopeTemp = Sigil<'%'>;
using AggregateType = Sigil<':'>;
```

Thus, template specialization is used instead.
There is some duplication, but it appears to be controllable:
```cpp
  void WriteLabel_(const qbe::user_defined::BlockLabel& label) {
    Write_("{}\n", label);
  }
  void WriteLabel_(const qbe::compiler_generated::BlockLabel& label) {
    Write_("{}\n", label);
  }
```

Two namespaces are introduced to separate user-defined and compiler-generated sigils, and I have the compiler-generated sigils directly included in the namespace of the _qbe_ir_generator.cpp_ file since compiler-generated sigils are used frequently:

```cpp
using namespace qbe::compiler_generated;
namespace user_defined = qbe::user_defined;
```

Resolves: #123


src/qbe_ir_generator.cpp
```diff
@@ -407,42 +411,43 @@ void QbeIrGenerator::Visit(const SwitchStmtNode& switch_stmt) {
...
 void QbeIrGenerator::GenerateCases_(const SwitchStmtNode& switch_stmt) {
   auto this_switch_info = switch_infos.back();
-  this_switch_info->exit_label =
-      std::make_shared<BlockLabel>("switch_exit", NextLabelNum());
   label_views_of_jumpable_blocks.push_back(
-      {// The break statement only jumps to the exit label; the entry label is
-       // irrelevant.
-       .entry = "",
-       .exit = this_switch_info->exit_label->name()});
+      {// XXX: The break statement only jumps to the exit label;
```
leewei05, line 431: Maybe use `FIXME`?

Lai-YT: Sure. :ok_hand:

leewei05 (approved):
LGTM! 🚀

PR #127, opened 2024-04-03
Introduce concept of scope & Check whether label is defined

Lai-YT:
- Each scope now associates with a `ScopeKind`. Please refer to _include/scope.hpp_: `struct Scope`.
  - `ScopeStack::PushScope` and `ScopeStack::Add` now require a `ScopeKind`. Modifications have been made in `src/type_checker.cpp` to assign scope kinds to scopes and symbols.

- `ScopeStack` now supports merging the current scope with the next scope. This functionality is implemented in _include/scope.hpp_: `ScopeStack::MergeWithNextScope`.
  - This enhancement specifically addresses including function parameters in the scope of the function body.

- The checking for "Use of undefined label" and "redefinition of label" now utilizes an auxiliary map. This change is implemented in _src/type_checker.cpp_: `<unnamed>::label_defined`.
  - The undefined-check is performed just before the function scope ends.

> [!important]
> Currently, a function scope is opened by a function, but no label is actually added to such a scope (symbol table). This is due to the following reasons:
> 1. Labels reside in their own scope and can only be redefined by labels.
> 2. The redefinition check of a label is performed by looking up the auxiliary map instead of the scope stack. If we were to look up the scope stack, we would need to support looking up a certain kind of scope or a certain kind of symbol, to avoid mistakenly `goto` a variable or assign to a label.
>
> In fact, I also question the usefulness of scope kinds. Although they are defined in the manual, I have yet to discern their semantic differences.
> @leewei05, do you think we should still add the labels into the scope? Or should we ignore the concept of scope kind altogether? Or should we defer this decision and leave the labels as not added?

### DIFF
--- a/src/type_checker.cpp
+++ b/src/type_checker.cpp
@@ -182,7 +182,8 @@ void TypeChecker::Visit(ReturnStmtNode& ret_stmt) {
 void TypeChecker::Visit(GotoStmtNode& goto_stmt) {
   // NOTE: We can know whether a label is defined until the function is about to
   // end. Thus, it's checked in the function definition.
-  // Also the lookup from the environment is not necessary.
+  // Also the lookup from the environment is not necessary. In fact, labels are
+  // not added to the environment.
   const bool is_not_defined =
       label_defined.find(goto_stmt.label) == label_defined.end() ||
       !label_defined.at(goto_stmt.label);
@@ -229,10 +230,6 @@ void TypeChecker::Visit(SwitchStmtNode& switch_stmt) {
 }
 
 void TypeChecker::Visit(IdLabeledStmtNode& id_labeled_stmt) {
-  // FIXME: Labels have error type.
-  env_.Add(std::make_unique<SymbolEntry>(id_labeled_stmt.label,
-                                         PrimitiveType::kUnknown),
-           ScopeKind::kFunc);
   label_defined[id_labeled_stmt.label] = true;
   id_labeled_stmt.stmt->Accept(*this);
 }
--- a/src/type_checker.cpp
+++ b/src/type_checker.cpp
@@ -13,6 +13,7 @@
 
 #include "ast.hpp"
 #include "operator.hpp"
+#include "scope.hpp"
 #include "symbol.hpp"
 #include "type.hpp"
 
@@ -52,7 +53,8 @@ void TypeChecker::Visit(DeclNode& decl) {
     // TODO: redefinition of 'id'
   } else {
     auto symbol = std::make_unique<SymbolEntry>(decl.id, decl.type);
-    env_.Add(std::move(symbol));
+    // TODO: May be file scope once we support global variables.
+    env_.Add(std::move(symbol), ScopeKind::kBlock);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/type_checker.cpp
```diff
@@ -152,7 +180,16 @@ void TypeChecker::Visit(ReturnStmtNode& ret_stmt) {
...
 
 void TypeChecker::Visit(GotoStmtNode& goto_stmt) {
-  // FIXME: To allow forward reference, we assume the label is declared.
+  // NOTE: We can know whether a label is defined until the function is about to
+  // end. Thus, it's checked in the function definition.
+  // Also the lookup from the environment is not necessary. In fact, labels are
+  // not added to the environment.
+  const bool is_not_defined =
+      label_defined.find(goto_stmt.label) == label_defined.end() ||
+      !label_defined[goto_stmt.label];
```
leewei05, line 189: Use `at()` instead?
```
label_defined.at(goto_stmt.label); 
```

src/type_checker.cpp
```diff
@@ -193,12 +230,11 @@ void TypeChecker::Visit(SwitchStmtNode& switch_stmt) {
...
 
 void TypeChecker::Visit(IdLabeledStmtNode& id_labeled_stmt) {
-  // TODO: To allow forward reference, the label has to be added in some other
-  // way.
-  // TODO: The label has function scope.
-  // FIXME: Labels have error type.
-  env_.Add(std::make_unique<SymbolEntry>(id_labeled_stmt.label,
-                                         PrimitiveType::kUnknown));
+  if (label_defined.find(id_labeled_stmt.label) != label_defined.end() &&
+      label_defined[id_labeled_stmt.label]) {
```
leewei05, line 234: Use `at()` instead?
```
label_defined.at(id_labeled_stmt.label); 
```

leewei05:
> @leewei05, do you think we should still add the labels into the scope? Or should we ignore the concept of scope kind altogether? Or should we defer this decision and leave the labels as not added?

I think we can keep the concept of scope kind just for now. Couldn't think of certain use cases for utilizing different scope kind, but I think this information would be quite handy if we started to implemented global variables, struct etc.

As for whether we should add labels into the scope, I prefer not to store them into the scope. I think it is easier to implement with the current `label_defined` map. If there's a label that has the same name with an identifier, we will need extra work to check them. Like the following code, there will be two `i` with different scope kind where we will need to check, and we couldn't be sure whether the first `i` is a label or an identifier unless we check their scope kind.

```
#include <stdio.h>

int main() {
    int i = 5;
    goto j;
i:
    i = 4;
    goto end;
j:
    printf("%d\n", i);
    goto i;
    
end:
    printf("%d\n", i);
    return 0;
}
```

leewei05 (changes requested):
Looking good! Only a few request changes. Or is it better to access key directly instead of `at()`?

Lai-YT:
Conflict resolved by rebasing. :love_you_gesture:

leewei05 (approved):
LGTM! Thanks for the great work!

PR #130, opened 2024-04-14
Implement pointer code generation

leewei05:
This PR adds implementation for pointer code generation. QBE doesn't have pointer type, thus we use `l` type 64 bits data to store address. Implementation details are written in the comments.

Implements #116


src/qbe_ir_generator.cpp
```diff
@@ -123,13 +123,37 @@ auto
...
     decl.init->Accept(*this);
     int init_num = num_recorder.NumOfPrevExpr();
-    WriteInstr_("storew {}, {}", FuncScopeTemp{init_num},
-                FuncScopeTemp{id_num});
+    // A pointer declaration may have two options for its right hand side:
+    if (decl.init->type.is_ptr) {
+      const auto* unary_expr = dynamic_cast<UnaryExprNode*>((decl.init).get());
+      // 1. int* a = &b; rhs is a reference of integer. We need to store b's
+      // address to a, where we need to map b's reg_num back to its id_num.
+      if (unary_expr != nullptr) {
```
Lai-YT, line 143: Since the use of `dynamic_cast` is only to test whether `decl.int` has the type of `UnaryExprNode*`, consider removing the intermediate variable `unary_expr`. At first glance, I thought it's going to be used somewhere else, as its lifetime is also unnecessarily long.
```suggestion
      // 1. int* a = &b; rhs is a reference of integer. We need to store b's
      // address to a, where we need to map b's reg_num back to its id_num.
      if (dynamic_cast<UnaryExprNode*>((decl.init).get())) {
```

Lai-YT (changes requested):
I will conduct a more thorough review later. :pray:

src/qbe_ir_generator.cpp
leewei05: Great suggestion! Fixed!

Lai-YT (approved):
This is a hard one. Thanks!

Issue #132, opened 2024-04-18
Support array feature


leewei05:
- [x] Parse array
- [x] Code generation array
- [ ] array initializer

PR #133, opened 2024-04-18
Support array declaration and array subscripting

leewei05:
- Change `DeclNode` to abstract struct and rename `DeclNode` to `DeclVarNode`.
- Create `DeclArrayNode` for supporting array.
- Create `ArrType`.
- Code generation for array declaration and array subscripting.
- Refactor allocation with `ToSize` function.

P.S. There is an unknown mystery [here](https://github.com/fruits-lab/VitaminC/pull/133/files#diff-20a77853f42f5ce8b91d36a6eb77c0e1bfc03203b6a98a9864bf49535ed69173R581). I couldn't find related information in QBE or cproc's docs, but it is necessary for now because it would print random numbers (I guess is address number of some kind) if we didn't add `mul 4`. I guess it's related to word-addressing.

Related to #132


include/type.hpp
```diff
@@ -74,11 +80,39 @@ class PtrType : public Type {
   }
 
   bool IsEqual(const Type& that) const noexcept override;
+  std::size_t ToSize() const override;
   std::string ToString() const override;
   std::unique_ptr<Type> Clone() const override;
 
  private:
   std::unique_ptr<Type> base_type_;
+  const std::size_t size_ = 8;
```
Lai-YT, line 89: Consider having the `ToSize` function return a constant `8` directly instead of having a constant data member.

include/type.hpp
```diff
@@ -74,11 +80,39 @@ class PtrType : public Type {
...
+      const noexcept {
+    return *base_type_;
+  }
+
+  bool IsArr() const noexcept override {
+    return true;
+  }
+
+  bool IsEqual(const Type& that) const noexcept override;
+  std::size_t ToSize() const override;
```
Lai-YT, line 107: Since the size of an array is the size of its base type times the number of elements, which may not be intuitive, I think it would be nice if we add a function comment for it.
```suggestion
  /// @note The size of an array is (size of base type) * (number of elements).
  std::size_t ToSize() const override;
```

src/type.cpp
```diff
@@ -34,10 +43,39 @@ bool PtrType::IsEqual(const Type& that) const noexcept {
...
+  if (const auto* that_arr = dynamic_cast<const ArrType*>(&that)) {
+    return that_arr->base_type_->IsEqual(*base_type_);
+  } else if (const auto* that_prim = dynamic_cast<const PrimType*>(&that)) {
+    return that_prim->IsEqual(*base_type_);
+  }
+  return false;
+}
+
+std::size_t ArrType::ToSize() const {
+  return base_type_->ToSize();
```
Lai-YT, line 68: I believe the size of an array is the size of the entire array?

include/type.hpp
```diff
@@ -74,11 +80,39 @@ class PtrType : public Type {
...
   std::unique_ptr<Type> base_type_;
+  const std::size_t size_ = 8;
+};
+
+class ArrType : public Type {
+ public:
+  explicit ArrType(std::unique_ptr<Type> base_type, std::size_t len)
+      : base_type_{std::move(base_type)}, len_{len} {}
+
+  const Type& base_type()  // NOLINT(readability-identifier-naming)
```
Lai-YT, line 97: Would `element_type` be more expressive?

src/type.cpp
```diff
@@ -34,10 +43,39 @@ bool PtrType::IsEqual(const Type& that) const noexcept {
...
 }
+
+bool ArrType::IsEqual(const Type& that) const noexcept {
+  if (const auto* that_arr = dynamic_cast<const ArrType*>(&that)) {
+    return that_arr->base_type_->IsEqual(*base_type_);
+  } else if (const auto* that_prim = dynamic_cast<const PrimType*>(&that)) {
+    return that_prim->IsEqual(*base_type_);
+  }
+  return false;
+}
```
Lai-YT, line 65: I think we need some revision on this. As mentioned in the reference manual at section 6.7.5.2:

- For two array types to be **compatible**, both must have compatible element types. Additionally, if both size specifiers are present and are integer constant expressions, then both size specifiers must have the same constant value.

The term **compatible** in the manual is akin to **equal** in our implementation.

include/ast.hpp
```diff
@@ -351,6 +375,21 @@ struct ArgExprNode : public ExprNode {
...
+  ArrSubExprNode(Location loc, std::unique_ptr<ExprNode> postfix_expr,
+                 std::unique_ptr<ExprNode> index)
+      : ExprNode{loc},
+        postfix_expr{std::move(postfix_expr)},
+        index{std::move(index)} {}
+
+  void Accept(NonModifyingVisitor&) const override;
+  void Accept(ModifyingVisitor&) override;
+
+  std::unique_ptr<ExprNode> postfix_expr;
```
Lai-YT, line 389: I think `expr` suffices in this case; the term `postfix` confuses me. :thinking:

Lai-YT (changes requested):

Lai-YT (changes requested):
Thank you for your dedication and patience in making these revisions and accommodating my requests.🙏
Regarding the mystery you mentioned, I believe `4` represents the size of `int`. When manipulating addresses in arrays, the offsets are typically measured in bytes. You should be able to calculate the resulting address (`res_addr`) directly by adding the base address with the size of the array. Our `size()` function should already handle this. However, if needed, we can revisit this issue in the future. We've already put in a lot of effort on this pull request.

src/type.cpp
```diff
@@ -34,10 +43,43 @@ bool PtrType::IsEqual(const Type& that) const noexcept {
...
+
+/// @note The size of an array is (size of element type) * (number of elements).
+std::size_t ArrType::size() const {
+  return element_type_->size() * len_;
+}
+
+std::string ArrType::ToString() const {
+  return element_type_->ToString();
+}
```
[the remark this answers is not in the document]
Lai-YT: Just like pointers, you can have the array type handle itself:
```cpp
element_type_->ToString() + "[" + std::to_string(len_) + "]";
```

leewei05:
> Thank you for your dedication and patience in making these revisions and accommodating my requests.🙏 Regarding the mystery you mentioned, I believe `4` represents the size of `int`. When manipulating addresses in arrays, the offsets are typically measured in bytes. You should be able to calculate the resulting address (`res_addr`) directly by adding the base address with the size of the array. Our `size()` function should already handle this. However, if needed, we can revisit this issue in the future. We've already put in a lot of effort on this pull request.

All fixed! Please take another look, thanks!

Lai-YT (approved):

PR #134, opened 2024-04-18
Fix type size of dereferencing & segmentation fault when taking address

Lai-YT:
# What's the Problem?

This pull request addresses 3 issues:

1. The type checker enforces that the operand of the dereferencing operator (`*`) be an _lvalue_, however, according to the manual, it doesn't.
   
   > 6.5.3.2 Address and indirection operators
   > Constraints
   > ...
   > 2 The operand of the unary `*` operator shall have pointer type.
   
2. The size used when loading the value of a dereferencing pointer should depend on the type of the result of the expression, which may be _long_ if having pointer type (once we support pointer to pointer), _word_ otherwise.

3. Getting the address of an identifier by _loading_ `id_num` causes a segmentation fault. This is because `id_num` itself already represents the address of the identifier. A simple _copy_ suffices.

# How to Test?

Here's the test case which triggers the segmentation fault:

```c
int main() {
  int a = 3;
  int b = 5;
  int c = *&a + *&b;
  return c;
}
```

The diffing on the QBE IR before and after the fix is as the following:

```diff
data $__builtin_print_format = align 1 { b "%d\012\000", }
export
function w $main() {
@.start.1
@.body.1
	%.1 =l alloc4 4
	%.2 =w copy 3
	storew %.2, %.1
	%.3 =l alloc4 4
	%.4 =w copy 5
	storew %.4, %.3
	%.5 =l alloc4 4
	%.6 =w loadw %.1
+ 	%.7 =l copy %.1
-	%.7 =l loadl %.1
 	%.8 =w loadw %.7
	%.9 =w loadw %.3
+ 	%.10 =l copy %.3
-	%.10 =l loadl %.3
 	%.11 =w loadw %.10
	%.12 =w add %.8, %.11
	storew %.12, %.5
	%.13 =w loadw %.5
	ret %.13
}
```


leewei05 (approved):
LGTM! Pointer is really tricky! Thanks for the fix. 👍

PR #136, opened 2024-04-22
Generate executables directly

leewei05:
- Output option will be used to name executable instead of IR file. I recommend using the legendary `a.out` as default. 😎 
- Add `target` option to specify the target IR. We only support `qbe` for now, we will support `llvm` in the future.


Lai-YT (approved):
LGTM :love_you_gesture:

PR #138, opened 2024-04-26
Support pointer to pointers

Lai-YT:
- A type can now hold another type as its base type.
- Polymorphic cloning is used to clone non-copyable types.
- Types are compared using `IsEqual` instead of `operator==`.
- AST nodes hold types by holding a unique pointer.
- Convert AST nodes to string using `ToString`.

Related to #116.


leewei05 (approved):
LGTM! 🚀

PR #139, opened 2024-04-28
Support Array initialization

### LINKED ISSUE #132: Support array feature

leewei05:
- Array initialization code generation
- Fix array subscripting bug
- Group non-terminal expressions in yacc

Resolves #132


test/codegen/array.c
```diff
@@ -11,5 +11,12 @@ int main() {
   a[3] = 2 + a[2];
   __builtin_print(a[3]);
 
+  int b = 2;
+  int c = 0;
+  int d[4] = {1, b, c = 3, 4};
+  for (int i = 0; i < 4; i = i + 1) {
```
Lai-YT, line 17: Suggest not using loop here so that the test doesn't fail when the loop construct is broken.

src/qbe_ir_generator.cpp
```diff
@@ -151,11 +151,32 @@ void QbeIrGenerator::Visit(const DeclVarNode& decl) {
...
 
-void QbeIrGenerator::Visit(const DeclArrNode& array_decl) {
-  int id_num = NextLocalNum();
-  WriteInstr_("{} =l alloc{} {}", FuncScopeTemp{id_num},
-              array_decl.type->element_type().size(), array_decl.type->size());
-  id_to_num[array_decl.id] = id_num;
+void QbeIrGenerator::Visit(const DeclArrNode& arr_decl) {
+  int base_addr_num = NextLocalNum();
+  std::size_t element_size = arr_decl.type->element_type().size();
+  WriteInstr_("{} =l alloc{} {}", FuncScopeTemp{base_addr_num}, element_size,
```
Lai-YT, line 157: Since the variable `element_size` is only used once, and `arr_decl.type->element_type().size()` is expressive enough, consider inline the variable:
```suggestion
  std::size_t element_size = arr_decl.type->element_type().size();
  WriteInstr_("{} =l alloc{} {}", FuncScopeTemp{base_addr_num}, arr_decl.type->element_type().size(),
```

src/qbe_ir_generator.cpp
```diff
@@ -151,11 +151,32 @@ void QbeIrGenerator::Visit(const DeclVarNode& decl) {
...
+              arr_decl.type->size());
+  id_to_num[arr_decl.id] = base_addr_num;
+
+  std::size_t index = 0;
+  for (auto& arr_init : arr_decl.init_list) {
+    arr_init->Accept(*this);
```
Lai-YT, line 163: Since we're iterating over `index` and `arr_decl.init_list` concurrently, it would be better to specify the index in the loop header and access `arr_decl.init_list` inside the loop. This approach ensures that the lifetime of `index` remains within the loop and eliminates the need for manual increments.
```suggestion
  for (auto i = std::size_t{0}, e = arr_decl.init_list->size(); i < e; ++i) {
    auto& arr_init = arr_decl.init_list.at(i);
    arr_init->Accept(*this);
```

Lai-YT (changes requested):
The formatting doesn't look right to me. Upon manual inspection of those lines, I discovered that they already fell within the 80-character line length limit. Could you please let me know which command and configuration you used to execute the formatter?

leewei05:
> The formatting doesn't look right to me. Upon manual inspection of those lines, I discovered that they already fell within the 80-character line length limit. Could you please let me know which command and configuration you used to execute the formatter?

Didn't change any configuration, just plain old `clang-format` inplace.

```
lee@ubuntu /U/l/d/VitaminC (array-initialier)> clang-format -i src/*.cpp
lee@ubuntu /U/l/d/VitaminC (array-initialier)> clang-format --version
Ubuntu clang-format version 15.0.7
```

Lai-YT (approved):
LGTM :hugs:

PR #142, opened 2024-04-30
Rewrite grammar of declaration to support complex types

Lai-YT:
This pull request overhauls the grammar of declarations to accommodate complex types, such as function pointers. Additionally, the new grammar, being a proper subset of the one defined in the reference manual, facilitates the introduction of new kinds of declarations, such as `struct` and `union`.

## Grammar

To illustrate how a function definition is parsed, let's consider the example `int *a(int *x)`:
- `int`: The return type is parsed as the non-terminal `declarator_specifiers`.
  - The `declarator_specifiers` derives to `type_specifier`, which ultimately stops at the terminal `INT`.
- `*a(int *x)`: The `*`, function name, and the parameters are parsed as the non-terminal `declarator`.
  - The `*` is absorbed by the non-terminal `pointer_opt`, leaving the remaining `a(int *x)` to be parsed as `direct_declarator`.
    - The `direct_declarator` is recursively defined; it recognizes the `ID`, leaving `(int *x)` to be handled by the rule `( parameter_type_list_opt )`.
      - `parameter_type_list_opt` recognizes `int *x` in an identical manner.
  
Key Observations:
- The pointer `*` is associated with the identifier (`declarator`), not the type specifier (`declarator_specifiers`).
- A type is parsed by multiple non-terminals. This enables parsing of function types as well as function pointers, where the type surrounds the identifier.

> [!note]
> There's a shift/reduce conflict that I have not yet figured out how to resolve. :disappointed:.

### Semantic Actions

As a declaration and its type are parsed step by step across several non-terminals, they cannot be fully resolved in a single semantic action. In cases where resolution is not yet possible, `nullptr` or `PrimitiveType::kUnknown` is passed to the constructors.

## AST Node

Given that function definitions and parameters are parsed as declarations, they inherit from `DeclNode`. This is reasonable since they both introduce identifiers with associated types. The function definition has its type set as `FuncType`, which includes the return type of the function and the types of its parameters. Notably, the function definition now stores the types of its parameters both at the parameter nodes and its type node.

---

The changes introduced by this pull request are intricate but crucial to understand. @leewei05, if you encounter any issues during the review or need clarification, please leave comments. I am also available for further discussion in person.

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -98,7 +98,8 @@ std::unique_ptr<Type> ResolveTypeAs(std::unique_ptr<Type> type,
 %nterm <std::unique_ptr<Type>> type_specifier
 // The followings also construct types, but they are not yet fully resolved.
 %nterm <std::unique_ptr<Type>> declaration_specifiers
-%nterm <std::vector<bool>> pointer_opt pointer
+// The number of '*'s.
+%nterm <int> pointer_opt pointer
 // The initializer of a simple variable is an expression, whereas that of an array or complex object is a list of expressions.
 %nterm <std::variant<std::unique_ptr<ExprNode>, std::vector<std::unique_ptr<ExprNode>>>> initializer
 %nterm <std::vector<std::unique_ptr<ExprNode>>> initializer_list
@@ -314,19 +315,7 @@ primary_expr: ID { $$ = std::make_unique<IdExprNode>(Loc(@1), $1); }
 /* TODO: If the init declarator doesn't present, e.g., `int;`, the declaration is still valid. */
 decl: declaration_specifiers init_declarator SEMICOLON {
     auto decl = $2;
-    auto type = $1;
-    // Cast the decl to its dynamic type to access the type.
-    // XXX: Consider placing the type under the base class.
-    if (dynamic_cast<DeclVarNode*>(decl.get())) {
-      auto decl_var = static_cast<DeclVarNode*>(decl.get());
-      type = ResolveTypeAs(std::move(decl_var->type), std::move(type));
-      decl_var->type = std::move(type);
-    } else if (dynamic_cast<DeclArrNode*>(decl.get())) {
-      auto decl_arr = static_cast<DeclArrNode*>(decl.get());
-      type = ResolveTypeAs(std::move(decl_arr->type), std::move(type));
-      // Since the array declaration node holds a array type, we need to cast it.
-      decl_arr->type = std::unique_ptr<ArrType>(static_cast<ArrType*>(type.release()));
-    }
+    decl->type = ResolveTypeAs(std::move(decl->type), $1);
     $$ = std::move(decl);
   }
   ;
@@ -335,8 +324,7 @@ decl: declaration_specifiers init_declarator SEMICOLON {
 /* TODO: storage class specifier, type qualifier, function specifier */
 declaration_specifiers: type_specifier declaration_specifiers {
     // The declaration specifier wraps the type specifier.
-    auto type = ResolveTypeAs($1, $2);
-    $$ = std::move(type);
+    $$ = ResolveTypeAs($1, $2);
   }
   | type_specifier { $$ = $1; }
   ;
@@ -348,11 +336,15 @@ init_declarator: declarator { $$ = $1; }
     auto decl = $1;
     auto init = $3;
     if (std::holds_alternative<std::unique_ptr<ExprNode>>(init)) {
+      auto* var_decl = dynamic_cast<DeclVarNode*>(decl.get());
+      assert(var_decl);
       auto init_expr = std::move(std::get<std::unique_ptr<ExprNode>>(init));
-      static_cast<DeclVarNode*>(decl.get())->init = std::move(init_expr);
+      var_decl->init = std::move(init_expr);
     } else { // The initializer is a list of expressions.
+      auto* arr_decl = dynamic_cast<DeclArrNode*>(decl.get());
+      assert(arr_decl);
       auto init_expr_list = std::move(std::get<std::vector<std::unique_ptr<ExprNode>>>(init));
-      static_cast<DeclArrNode*>(decl.get())->init_list = std::move(init_expr_list);
+      arr_decl->init_list = std::move(init_expr_list);
     }
     $$ = std::move(decl);
   }
@@ -378,21 +370,9 @@ type_specifier: INT { $$ = std::make_unique<PrimType>(PrimitiveType::kInt); }
 declarator: pointer_opt direct_declarator {
     @$ = @2; // Set the location to the identifier.
     auto declarator = $2;
-    for (auto is_pointer : $1) {
-      (void)is_pointer;
+    for (int i = 0, e = $1; i < e; ++i) {
       auto unknown_ptr_type = std::make_unique<PtrType>(std::make_unique<PrimType>(PrimitiveType::kUnknown));
-      // Cast the declarator to its dynamic type to access the type.
-      // XXX: Consider placing the type under the base class.
-      if (dynamic_cast<DeclVarNode*>(declarator.get())) {
-        auto decl_var = static_cast<DeclVarNode*>(declarator.get());
-        auto type = ResolveTypeAs(std::move(decl_var->type), std::move(unknown_ptr_type));
-        decl_var->type = std::move(type);
-      } else if (dynamic_cast<DeclArrNode*>(declarator.get())) {
-        auto decl_arr = static_cast<DeclArrNode*>(declarator.get());
-        auto type = ResolveTypeAs(std::move(decl_arr->type), std::move(unknown_ptr_type));
-        // Since the array declaration node holds a array type, we need to cast it.
-        decl_arr->type = std::unique_ptr<ArrType>(static_cast<ArrType*>(type.release()));
-      }
+      declarator->type = ResolveTypeAs(std::move(declarator->type), std::move(unknown_ptr_type));
     }
     $$ = std::move(declarator);
   }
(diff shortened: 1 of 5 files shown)


parser.y
```diff
@@ -10,13 +11,21 @@
 #include "operator.hpp"
 #include "type.hpp"
 
+namespace {
+/// @brief Resolved the innermost unknown type to the given type.
+/// @note If the type is a primitive type, the resolved type is returned.
```
leewei05, line 16: This comment is different from the implementation, where `resolved_type` is returned, only if `type` is a primitive `unknown type`. If `type` is a `PrimType::int`, do you intend to return `resolved_type`?

parser.y
```diff
@@ -331,19 +311,219 @@ primary_expr: ID { $$ = std::make_unique<IdExprNode>(Loc(@1), $1); }
...
   std::cerr << loc << ": " << err << std::endl;
 }
+
+namespace {
+
+std::unique_ptr<Type> ResolveTypeAs(std::unique_ptr<Type> type,
+                                    std::unique_ptr<Type> resolved_type) {
+  // Base case: this type itself is the unknown type to resolve.
+  if (type->IsPrim()) {
+    assert(type->IsEqual(PrimitiveType::kUnknown));
```
leewei05, line 512: Like I mentioned, what if `type` is `PrimitiveType::kInt`?  Return `nullptr`?

parser.y
```diff
@@ -331,19 +311,219 @@ primary_expr: ID { $$ = std::make_unique<IdExprNode>(Loc(@1), $1); }
...
 epsilon: %empty;
 %%
 
 void yy::parser::error(const yy::location& loc, const std::string& err) {
   std::cerr << loc << ": " << err << std::endl;
 }
+
+namespace {
+
+std::unique_ptr<Type> ResolveTypeAs(std::unique_ptr<Type> type,
```
leewei05, line 508: Can you give a simple example for this function?

leewei05 (commented):
I see you added a lot of TODOs for supporting new features, which is great. However, I want you to know that I am not interested in supporting some of the features, such as `storage class`, `type qualifier` and cases like `int;`. I prefer to finish basic feature (struct, union, global variables) in the least amount of time and move on to llvm backend support. For me, those features are worth `knowing` but not `implementing` right now.

parser.y
```diff
@@ -80,16 +89,24 @@
...
+%nterm <std::vector<std::unique_ptr<ParamNode>>> parameter_type_list_opt parameter_type_list parameter_list
+// The followings also declare an identifier, however, their types are not yet fully resolved.
+%nterm <std::unique_ptr<DeclNode>> declarator direct_declarator init_declarator
+%nterm <std::unique_ptr<Type>> type_specifier
+// The followings also construct types, but they are not yet fully resolved.
+%nterm <std::unique_ptr<Type>> declaration_specifiers
+// The number of '*'s.
+%nterm <int> pointer_opt pointer
+// std::variant is used to distinguish between a single expression and a list of expressions;
+// a list of a single expression is not the same as a single expression.
```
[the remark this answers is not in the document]
Lai-YT: Sure!

src/type.cpp
```diff
@@ -83,3 +83,41 @@ std::unique_ptr<Type> ArrType::Clone() const {
...
+  return false;
+}
+
+std::size_t FuncType::size() const {
+  const auto pointer_size = 8;
+  return pointer_size;
+}
+
+std::string FuncType::ToString() const {
+  // FIXME: This is to not break the tests; should include the parameter types.
```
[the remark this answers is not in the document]
Lai-YT: Currently, our _ast dump_ only displays the _return type_ of the function, but it would be more comprehensive to include the _parameter types_ as well.

Consider the following program as an example:
```c
int main(int argc, char** argv) {
  return 0;
}
```

LLVM generates the following dump:
```
...
`-FunctionDecl 0x11e3340 <test.c:2:1, line:4:1> line:2:5 main 'int (int, char **)'
  |-ParmVarDecl 0x11e31b8 <col:10, col:14> col:14 argc 'int'
  |-ParmVarDecl 0x11e3260 <col:20, col:27> col:27 argv 'char **'
  `-CompoundStmt 0x11e3468 <col:33, line:4:1>
    `-ReturnStmt 0x11e3458 <line:3:3, col:10>
      `-IntegerLiteral 0x11e3438 <col:10> 'int' 0
...
```

parser.y
```diff
@@ -138,32 +155,28 @@ func_def_list_opt: func_def_list_opt func_def {
...
+/* NOTE: The obsolete form of function definition is not supported,
+   e.g., `int max(a, b) int a, b; { return a > b ? a : b; }`. */
+func_def: declaration_specifiers declarator compound_stmt {
+    // The declarator shall already be a function declarator.
+    // We resolve its return type with the declaration specifiers and set the body.
+    auto func_def = $2;
+    assert(dynamic_cast<FuncDefNode*>(func_def.get()));
+    assert(func_def->type->IsFunc());
+    const auto* func_type = static_cast<FuncType*>(func_def->type.get());
+    auto resolved_return_type = ResolveTypeAs(func_type->return_type().Clone(), $1);
```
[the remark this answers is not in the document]
Lai-YT: The return type may be some pointer-ish type, as well as function pointers. This makes the underlying primitive type hidden in the nested type structure. Thus, we still need to _resolve_ it.

parser.y
Lai-YT: If type is a primitive type to int, the resolve function should not even be called, which is why I include an assertion in the beginning of the implementation.

parser.y
Lai-YT: If the primitive type is int, it is already resolved. Calling resolve on it should indicate an internal error has occurred.

parser.y
Lai-YT: Let's take a complex declaration as an example:
```c
int* ptr_arr[10];
```

- The declaration is recognized by the rule `decl: declaration_specifiers declarator SEMICOLON` (omit `init` for simplicity).
  - `int` is recognized by the non-terminal `declaration_specifier`, which constructs _an int type_ in its action.
  - `* ptr_arr[10]` is recognized by the rule `declarator: pointer_opt direct_declarator`:
    - The `pointer_opt` has its semantic value as `1`.
    - The `direct_declarator` has its semantic value as _an array to an unknown type_.
    - So in the following semantic action, we try to combine the two types into a single type: _an array to a pointer to an unknown type_.
      ```yacc
      declarator: pointer_opt direct_declarator {
          @$ = @2; // Set the location to the identifier.
          auto declarator = $2;
          for (int i = 0, e = $1; i < e; ++i) {
            auto unknown_ptr_type = std::make_unique<PtrType>(std::make_unique<PrimType>(PrimitiveType::kUnknown));
            declarator->type = ResolveTypeAs(std::move(declarator->type), std::move(unknown_ptr_type));
          }
          $$ = std::move(declarator);
        }
        ;
      ```
      To achieve this, we have to place the type _a pointer to an unknown type_ as the innermost type of _an array to a pointer to an unknown type_, which is what the function `ResolveTypeAs` does.
  - So in the following semantic action, we try to combine the type _an int type_ with the type _an array to a pointer to an unknown type_:
    ```yacc
    decl: declaration_specifiers init_declarator SEMICOLON {
        auto decl = $2;
        decl->type = ResolveTypeAs(std::move(decl->type), $1);
        $$ = std::move(decl);
      }
      ;
    ```
    This is achieved by placing _an int type_ as the innermost type of _an array to a pointer to an unknown type_.

The `ResolveTypeAs` function is implemented as a recursive function. It gets directly to the innermost type, using the `resolved_type` as the innermost type and wraps it with the layers of `type`. Take the above _an array to a pointer to an unknown type_ resolution as an example:
- The outermost type is an array, thus the following condition is entered:
    ```cpp
      if (type->IsArr()) {
        auto arr_type = static_cast<ArrType*>(type.get());
        auto type_ = ResolveTypeAs(arr_type->element_type().Clone(), std::move(resolved_type));
        return std::make_unique<ArrType>(std::move(type_), arr_type->len());
      }
    ```
    It recursively tries to resolve its element type _a pointer to an unknown type_.
    - The type is now a pointer, thus the following condition is entered:
      ```cpp
        if (type->IsPtr()) {
          auto ptr_type = static_cast<PtrType*>(type.get());
          auto type_ = ResolveTypeAs(ptr_type->base_type().Clone(), std::move(resolved_type));
          return std::make_unique<PtrType>(std::move(type_));
        }
      ```
      It tries to resolve its base type, which is _an unknown type_.
      - Finally, the type is a primitive type _an unknown type_. The base condition is entered:
        ```cpp
        if (type->IsPrim()) {
          return resolved_type;
        }
        ```
        The `resolved_type` is the _an int type_ in this case.
- Now, we go all the way up.
  - `return std::make_unique<PtrType>(std::move(type_ /* an int type */));`
  - `return std::make_unique<ArryType>(std::move(type_ /* a pointer to an int type */));`
  Getting the resolved type _an array to a pointer to an int type_.

I noticed that in the assignment `auto type_ = ResolveTypeAs(arr_type->element_type().Clone(), std::move(resolved_type));`, the variable `type_` should better be named as `resolved_type`, as the type it holds is resolved. How do you think about this? The parameter `resolved_type` can be reassigned in this case. The following is the concept code:
```cpp
  if (type->IsPtr()) {
    auto ptr_type = static_cast<PtrType*>(type.get());
    resolved_type = ResolveTypeAs(ptr_type->base_type().Clone(), std::move(resolved_type));
    return std::make_unique<PtrType>(std::move(resolved_type));
  }
```

parser.y
```diff
@@ -138,32 +155,28 @@ func_def_list_opt: func_def_list_opt func_def {
...
-    $$ = std::vector<std::unique_ptr<ParamNode>>{};
-    $$.push_back($1);
-  }
-  ;
-
-parameter: type_specifier ID {
-    $$ = std::make_unique<ParamNode>(Loc(@2), $2, $1);
+/* 6.9.1 Function definitions */
+/* NOTE: The obsolete form of function definition is not supported,
+   e.g., `int max(a, b) int a, b; { return a > b ? a : b; }`. */
```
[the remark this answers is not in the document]
Lai-YT: I left this comment as a _NOTE_ to remind us (and possibly users) that we do not support it. If we are to write some documentation for _VitaminC_, we may want to list some supported and unsupported features, and this comment can serve as a reminder.

Lai-YT:
> I see you added a lot of TODOs for supporting new features, which is great. However, I want you to know that I am not interested in supporting some of the features, such as `storage class`, `type qualifier` and cases like `int;`. I prefer to finish basic feature (struct, union, global variables) in the least amount of time and move on to llvm backend support. For me, those features are worth `knowing` but not `implementing` right now.

I understand that these features are more like additions rather than essential requirements for building a usable language. I left these comments to remind us of what else we would need if we were to revisit supporting more _C_ language features. Since the grammar itself is not trivial and we may be less familiar with it at that time, I would like to leave some _TODO_s. These can also be listed as unsupported features for users to be aware of. It's very difficult to determine what is supported and what is not just by looking at the grammar or other analysis components.

parser.y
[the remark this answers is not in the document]
leewei05: Fine by me 👌

src/type.cpp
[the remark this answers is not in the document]
leewei05: Noted!

parser.y
leewei05: > How do you think about this? The parameter resolved_type can be reassigned in this case. The following is the concept code:

Sure! I second this.

leewei05 (approved):
LGTM! Thank you for your effort! Tai Shen La! 🥳

PR #144, opened 2024-05-04
Support Postfix increment and decrement

leewei05:
- Postfix increment and decrement expression

P.S: Pointer arithmetic is not yet supported.


Lai-YT (changes requested):
Nice! :pray: Just a small suggestion on the comment. Everything else looks great to me. :love_you_gesture:

Lai-YT (approved):

PR #145, opened 2024-05-04
Refactor expressions

leewei05:
As I was implementing `struct` and `union`, I found the use for `const_expr`, so I might as well refactor expression grammar to match the C standard. Also, we can remove precedence of arithmetic operators, since the grammar has taken care of it.

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -78,7 +78,7 @@ std::unique_ptr<Type> ResolveType(std::unique_ptr<Type> resolved_type,
 %define api.location.file none
 
 %token MINUS PLUS STAR DIV MOD ASSIGN
-%token EXCLAMATION TILDE AMPERSAND
+%token EXCLAMATION TILDE AMPERSAND QUESTION
 %token COMMA SEMICOLON COLON
 // (), {}, []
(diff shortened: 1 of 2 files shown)


parser.y
```diff
@@ -286,13 +278,83 @@ unary_expr: postfix_expr { $$ = $1; }
...
+cond_expr: logic_or_expr
+  | logic_or_expr QUESTION expr : cond_expr
+  ;
+
+/* 6.5.16 Assignment operators */
+/* TODO: support multiple assignment operators */
+assign_expr: unary_expr ASSIGN expr {
+    $$ = std::make_unique<SimpleAssignmentExprNode>(Loc(@2), $1, $3);
+  }
+  ;
```
Lai-YT, line 354: I think it's fine for us to complete the grammar at this point:

```suggestion
assign_expr: unary_expr ASSIGN assign_expr {
    $$ = std::make_unique<SimpleAssignmentExprNode>(Loc(@2), $1, $3);
  }
  | cond_expr { $$ = $1; }
  ;
```

Lai-YT (changes requested):

parser.y
leewei05: Thanks! Missed that one 😅

Lai-YT (approved):
LGTM 🎉

PR #147, opened 2024-05-08
Support function pointers

Lai-YT:
During the implementation of function pointers, several important concepts and a new type comparison mechanism have been introduced:

- A function can be assigned to a function pointer, which is checked by `Type::IsConvertibleTo`.
- The result of the `IdExprNode` of a function is the address of the function.
- Taking the address of a function is a no-op, as the function itself already evaluates to its address; likewise, dereferencing a function pointer is also a no-op.
- Functions, except built-ins, are called through a register that holds their address.
  ```
  %.1 =l copy $func
  %.2 =w call %.1()
  ```
- The type size is `l` when attempting to load a function type.

Please take a look at the test cases to see what we can achieve now! :muscle:

Related to #116.


leewei05 (approved):
LGTM!

PR #151, opened 2024-05-13
Implement bitwise shift operation

leewei05:
- Fix clang-tidy by adding missing header.
- Implement bitwise shift left and shift right.

> The result of E1 >> E2 is E1 right-shifted E2 bit positions. If E1 has an unsigned type
or if E1 has a signed type and a nonnegative value, the value of the result is the integral
part of the quotient of E1 / 2E2. If E1 has a signed type and a negative value, the
resulting value is implementation-defined.

We only support signed integer for now, so nonnegative value should perform **arithmetic shift right**. Since right shift negative signed type is compiler defined, I suggest using the same operation `sar` for it.


src/qbe_ir_generator.cpp
```diff
@@ -70,6 +70,12 @@ std::string GetBinaryOperator(BinaryOperator op) {
       return "xor";
     case BinaryOperator::kOr:
       return "or";
+    case BinaryOperator::kShl:
+      return "shl";
+    // Arithmetic shift right (sar) is only equivalent to a division by a power
+    // of two for non-negative numbers.
```
Lai-YT, line 76: Suggest making this a `NOTE:` since this is an important detail. State that _right-shifting negative numbers is implementation-defined; we choose to also use arithmetic shift on them._

Lai-YT (changes requested):

Lai-YT (approved):
LGTM :nerd_face:

PR #154, opened 2024-05-19
Logical expression implementation

leewei05:
Since QBE doesn't have logical operation instructions, we need to implement logical expressions with comparison and jump instructions.


src/qbe_ir_generator.cpp
```diff
@@ -822,14 +822,49 @@ void QbeIrGenerator::Visit(const BinaryExprNode& bin_expr) {
...
-  // 2. QBE base data type 'w' | 'l' | 's' | 'd': currently only supports 'w'.
-  WriteInstr_("{} =w {} {}, {}", FuncScopeTemp{num},
-              GetBinaryOperator(bin_expr.op), FuncScopeTemp{left_num},
-              FuncScopeTemp{right_num});
-  num_recorder.Record(num);
+  // The && operator shall yield 1 if both of its operands compare unequal to
+  // 0; otherwise, it yields 0; The || operator shall yield 1 if either of its
+  // operands compare unequal to 0; otherwise, it yields 0.
+  if (bin_expr.op == BinaryOperator::kLand ||
+      bin_expr.op == BinaryOperator::kLor) {
```
Lai-YT, line 829: How about placing an explanation of why we need this extra condition above the condition itself and moving the implementation details inside the loop?

```suggestion
  // Due to the lack of direct support for logical operators in QBE, we
  // implement logical expressions using comparison and jump instructions.
  if (bin_expr.op == BinaryOperator::kLand ||
      bin_expr.op == BinaryOperator::kLor) {
     // The && operator yields 1 if both of its operands compare unequal to 0;
     // otherwise, it yields 0. The || operator yields 1 if either of its
     // operands compare unequal to 0; otherwise, it yields 0.
```

src/qbe_ir_generator.cpp
```diff
@@ -822,14 +822,49 @@ void QbeIrGenerator::Visit(const BinaryExprNode& bin_expr) {
...
+    const int label_num = NextLabelNum();
+    auto rhs_label = BlockLabel{"logic_rhs", label_num};
+    // Early exit after evaluating the first operand.
+    auto exit_label = BlockLabel{"logic_exit", label_num};
```
Lai-YT, line 833: The early exit is commonly referred to as [short-circuit evaluation](https://en.wikipedia.org/wiki/Short-circuit_evaluation). How about we name the label accordingly to make it more easily understood?

```suggestion
    // Early exit after evaluating the first operand.
    auto short_circuit_label = BlockLabel{"short_circuit", label_num};
```

test/codegen/bin_expr.c
```diff
@@ -7,5 +7,9 @@ int main() {
   __builtin_print(-1 << 31);
   __builtin_print(65536 >> 16);
   __builtin_print(-1 >> 16);
+  __builtin_print(1 && 1);
+  __builtin_print(0 && 1);
+  __builtin_print(1 || 0);
+  __builtin_print(0 || 0);
```
Lai-YT, line 13: How about representing `true` with a value other than `1`, so that if we accidentally only treat `1` as `true`, this test will fail?

Lai-YT (changes requested):

Lai-YT (approved):
LGTM! Thanks for building on this!

PR #155, opened 2024-05-20
Add declaration type in scope stack

leewei05:
Similar with identifiers, declaration type also has scope. The PR supports scope for types.

- Add `DeclTypeEntry` and `DeclTypeTable` for record declaration, also for future `typedef` and `enum` support.
- Since declaration shares the same operations `Add` and `Probe` for their tables, I used function template for reducing duplicate code. With suffix `Symbol` or `Type` in their method names, it will update symbol table or type table accordingly. Let me know if you prefer other solutions.
- Remove `symbol.cpp` because I moved its function template to header. This is to avoid undefined reference in linking stage since it is not included in the compilation process. I think this is easier than `explicit instantiate template`.


include/symbol.hpp
```diff
@@ -16,18 +16,42 @@ struct SymbolEntry {
       : id{std::move(id)}, type{std::move(expr_type)} {}
 };
 
-class SymbolTable {
+struct DeclTypeEntry {
```
Lai-YT, line 19: How about simply `TypeEntry`? Since both the symbol and type are declared, I think we don't need to encode this in the name.

include/scope.hpp
```diff
@@ -27,57 +27,49 @@ enum class ScopeKind : std::uint8_t {
...
+        return ((*it).*table)->Add(std::move(entry));
       }
     }
     throw NotInSuchKindOfScopeError{""};
   }
 
   /// @brief Looks up the `id` from through all scopes.
   /// @throws `NotInScopeError`
-  std::shared_ptr<SymbolEntry> LookUp(const std::string& id) const {
+  template <typename Table, typename EntryType>
```
Lai-YT, line 66: Can simplify to `Entry`. It's obvious that this is a type.
```suggestion
  template <typename Table, typename Entry>
```

include/symbol.hpp
```diff
@@ -16,18 +16,42 @@ struct SymbolEntry {
...
+    return nullptr;
+  }
 
  private:
-  std::map<std::string, std::shared_ptr<SymbolEntry>> entries_{};
+  std::map<std::string, std::shared_ptr<EntryType>> entries_{};
 };
 
+/// @brief Stores identifiers.
+using SymbolTable = TableTemplate<SymbolEntry>;
```
Lai-YT, line 53: I'm a bit confused about the difference between _identifier_ and _symbol_. The `DeclTypeEntry` also holds a `std::string id`, so I believe `id` is a more general concept. How about we have _symbol_ refer to variable ids and function ids? 

As a result, `DeclType` can naturally refer to the ids that declare a type. This way, the ambiguous comment can be removed.

```suggestion
using SymbolTable = TableTemplate<SymbolEntry>;
```

Lai-YT (commented):
1. Since the type of `Table` in use is closely related to the type of `Entry`, it's unlikely that users will define their own tables beyond the two we currently have. Therefore, we can utilize _explicit template instantiation_, which has the advantage of removing dependencies, reducing compile time, and minimizing code size. :fire: 
2. Although the suffix `Symbol` is redundant in the use case of `Add` (e.g., `env_.AddSymbol(std::move(symbol), ScopeKind::kBlock);`) and we could use function overloading due to distinct parameter types, we cannot remove the suffix from the `Probe` function as it cannot be overloaded. Therefore, I think it is fine to keep the function name `AddSymbol` verbose to align with `ProbeSymbol`. :ok_man:

include/scope.hpp
```diff
@@ -115,6 +105,39 @@ class ScopeStack {
...
+  /// @brief Adds the `entry` to the top-most scope of the `kind`.
+  /// @tparam Table The type of the table to add the entry.
+  /// @tparam Entry The type of the entry to add.
+  /// @param kind The kind of the scope to add the entry.
+  /// @param table The class member pointer to the table to add the entry.
+  /// @return The added entry if the `id` of the `entry` isn't already in such
+  /// scope; otherwise, the original entry.
+  template <typename Table, typename Entry>
+  std::shared_ptr<Entry> AddEntry_(std::unique_ptr<Entry> entry, ScopeKind kind,
+                                   std::unique_ptr<Table> Scope::*table);
```
Lai-YT, line 118: The template parameter `Table` can be deduced from `Entry`, as it is actually `TableTemplate<Entry>`. I suggest we unify them. 🤓

include/scope.hpp
```diff
@@ -86,22 +78,30 @@ class ScopeStack {
 
   /// @brief Probes the `id` from the top-most scope.
   /// @throws `NotInScopeError`
-  std::shared_ptr<SymbolEntry> Probe(const std::string& id) const {
+  template <typename Table, typename EntryType>
+  std::shared_ptr<EntryType> ProbeEntry(
+      const std::string& id, std::unique_ptr<Table> Scope::*table) const {
```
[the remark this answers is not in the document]
Lai-YT: Oops! I misunderstood what the parameter represents. Since it specifically denotes the class member pointer, it's fine to keep it as is!

src/scope.cpp
```diff
@@ -0,0 +1,119 @@
...
+ScopeStack::LookUpEntry_<TypeTable, TypeEntry>(
+    const std::string&, std::unique_ptr<TypeTable> Scope::*) const;
+
+template std::shared_ptr<SymbolEntry>
+ScopeStack::ProbeEntry_<SymbolTable, SymbolEntry>(
+    const std::string&, std::unique_ptr<SymbolTable> Scope::*) const;
+
+template std::shared_ptr<TypeEntry>
+ScopeStack::ProbeEntry_<TypeTable, TypeEntry>(
+    const std::string&, std::unique_ptr<TypeTable> Scope::*) const;
```
Lai-YT, line 119: We don't need these explicit instantiations because they are already instantiated when defining the public functions (`AddSymbol`, `AddType`, etc.).

src/scope.cpp
```diff
@@ -0,0 +1,95 @@
+#include "scope.hpp"
+
+#include <cstdint>
```
Lai-YT, line 3: Unused include.

src/scope.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <memory>
+#include <stdexcept>
```
Lai-YT, line 5: Unused include.

Lai-YT (changes requested):

Lai-YT (approved):
Nice change using templates to support common functions! LGTM 🎉

PR #156, opened 2024-05-20
Implement conditional expression

leewei05:
The first operand is evaluated. The second operand is evaluated only if the first compares unequal to 0; the third operand is evaluated only if the first compares equal to 0.


src/qbe_ir_generator.cpp
```diff
@@ -653,6 +653,39 @@ void QbeIrGenerator::Visit(const ArrSubExprNode& arr_sub_expr) {
...
+  cond_expr.predicate->Accept(*this);
+  const int first_num = num_recorder.NumOfPrevExpr();
+  cond_expr.then->Accept(*this);
+  const int second_num = num_recorder.NumOfPrevExpr();
+  cond_expr.or_else->Accept(*this);
+  const int third_num = num_recorder.NumOfPrevExpr();
+  // The second operand is evaluated only if the first compares unequal to
+  // 0; the third operand is evaluated only if the first compares equal to
+  // 0; the result is the value of the second or third operand (whichever is
+  // evaluated).
```
Lai-YT, line 666: Seems like you're evaluating both operands before checking the predicate.

leewei05: Nice catch! 🦅

Lai-YT (approved):
LGTM 🥳

PR #158, opened 2024-06-02
Implement initializer for array, struct and union

leewei05:
The goal of this PR is to implement initializer for Record declaration types (struct and union). Since array also uses initializer to initialize its elements, it is also a part of this PR.

We can break this PR down into multiple parts:

### Initializer
- Update initializer grammar
- Create initializer AST node `InitExprNode`

### Designation (part of initializer)
- Update designation grammar
- Create a new abstract AST node `DesNode` for designator, and two concrete AST node `ArrDesNode` for array subscripting designator (`[0] = 1`, `[1].abc = 4`) and `IdDesNode` for identifier designator (`.date = 12`). A collection of these two nodes form a **designation** for a member.

### Struct, Union variable declaration
I decided to create a new node `RecordVarDeclNode` to separate from `VarDeclNode` for struct, union declaration because we will need to implement nested initializer in the future. We will only need to update `RecordVarDeclNode`, instead of changing` VarDeclNode`, which may break previous implementation and tests. In addition, `RecordVarDeclNode` can separate **type declaration** (declaration for a struct or an union type) and **variable declaration** (declaration for a variable with the type of a struct or an union).

### StructType, Union Type
- Create a new `id` member for both these two types for showing the type name in ast_dumper.


parser.y
```diff
@@ -396,6 +398,18 @@ decl: declaration_specifiers init_declarator_opt SEMICOLON {
...
+      if (auto* rec_var_decl = dynamic_cast<RecordVarDeclNode*>(init_decl.get())) {
+        // A struct or union variable.
+        rec_var_decl->type = ResolveType(std::move(rec_decl->type), std::move(rec_var_decl->type));
+        decl = std::move(init_decl);
+      } else if (auto* arr_decl = dynamic_cast<ArrDeclNode*>(init_decl.get())) {
+        // An array with struct or union elements.
+        arr_decl->type = ResolveType(std::move(rec_decl->type), std::move(arr_decl->type));
+        decl = std::move(init_decl);
+      }
+
```
Lai-YT, line 412: This code can be simplified by utilizing the inheritance hierarchy. Since the data member 'type' is held by the base class `DeclNode`, there is no need to cast to the concrete classes. 
```suggestion
      if (init_decl) {
        init_decl->type = ResolveType(std::move(rec_decl->type), std::move(init_decl->type));
        decl = std::move(init_decl);
      }
```
> [!IMPORTANT]
> Please notice the check `if (init_decl)`, since we are having unnamed struct/union in our test cases, the `init_declarator_opt` could be a null pointer. I kindly request that you solve the unnamed primitive types as also, which is included by `/* TODO: If the init declarator doesn't present, e.g., ``int;``, the declaration is still valid. */`, so that the `if (init_decl)` check can be moved one level above to cover both primitive and user defined types.

test/typecheck/struct.c
```diff
@@ -13,5 +13,15 @@ int main() {
...
 
+  struct birth bd1 = {
+    .date = 1,
+    .month = 1,
+    .year = 1995,
+  };
+
+  struct birth bd2 = {3, 3, 1998};
+
+  struct birth bd3[3] = { [0].date = 4, [1].year = 1999};
```
Lai-YT, line 24: Minor formatting issue; the `.exp` file may have to update accordingly.
```suggestion
  struct birth bd3[3] = {[0].date = 4, [1].year = 1999};
```

Lai-YT (changes requested):
The code generation seems to also be challenging. :sob:

parser.y
```diff
@@ -79,7 +79,7 @@ std::unique_ptr<Type> ResolveType(std::unique_ptr<Type> resolved_type,
 
 %token MINUS PLUS STAR DIV MOD ASSIGN
 %token EXCLAMATION TILDE AMPERSAND QUESTION
-%token COMMA SEMICOLON COLON
+%token COMMA PERIOD SEMICOLON COLON
```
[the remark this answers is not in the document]
leewei05: > How about `DOT`? 😄 I find it more natural to say `a.b` as `a dot b`. 🙋

Sure! I was thinking about changing it to `DOT`. 😂

include/ast.hpp
```diff
@@ -351,6 +381,42 @@ struct ExprStmtNode : public StmtNode {
   std::unique_ptr<ExprNode> expr;
 };
 
+/// @brief An initializer initializes a member in array, struct or union.
+struct InitExprNode : public ExprNode {
+  InitExprNode(Location loc, std::vector<std::unique_ptr<DesNode>> des,
+               std::unique_ptr<ExprNode> expr)
+      : ExprNode{loc}, des{std::move(des)}, expr{std::move(expr)} {}
```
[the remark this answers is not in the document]
leewei05: > To confirm my understanding, the designators can appear in a chained manner (e.g., `[0].circle = 1` in our test case). This is why they are represented as a vector. The value `1` is the expression, and this design could be extended in the future to support nested initializers. Is that correct?

Yes, you are correct! Nested initializers is a tricky case to implement. 😮‍💨

include/type.hpp
```diff
@@ -189,13 +191,16 @@ class StructType : public Type {
...
+  std::string id_;
   std::vector<std::unique_ptr<Type>> field_types_;
 };
 
 class UnionType : public Type {
  public:
-  explicit UnionType(std::vector<std::unique_ptr<Type>> field_types)
-      : field_types_{std::move(field_types)} {}
+  /// @param id The identifier of the union type. May be empty ("") for unnamed
+  /// structs.
```
Lai-YT, line 201: :smile: 
```suggestion
  /// unions.
```

Lai-YT (changes requested):

PR #160, opened 2024-06-10
Support comma operator

Lai-YT:
I considered adding a new node instead of supporting the comma operator as one of the binary operators, since it requires completely different type checking and code generation compared to other binary operators. This is evident from the ad-hoc rules added to both the type checker and code generator. However, I decided to keep it as a binary operator to align with the fact that it is, indeed, a binary operator, similar to how Clang handles it.

Let me know your thoughts on this approach!


leewei05 (approved):
LGTM!

PR #162, opened 2024-06-16
Fix missing record declaration tests

leewei05:
My bad, missed this update on previous reviews. The current implementation only consider the case for `struct` and `union` variable initialization.

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -395,6 +395,11 @@ decl: declaration_specifiers init_declarator_list_opt SEMICOLON {
     auto decl_list = std::vector<std::unique_ptr<DeclNode>>{};
     if (std::holds_alternative<std::unique_ptr<Type>>(decl_specifiers)) {
       auto type = std::move(std::get<std::unique_ptr<Type>>(decl_specifiers));
+      if (init_decl_list.empty()) {
+        // A stand-alone type that doesn't declare any identifier, e.g., `int;`.
+        decl_list.push_back(std::make_unique<VarDeclNode>(Loc(@1), "", type->Clone()));
+      }
+
       for (auto& init_decl : init_decl_list) {
         if (init_decl) {
           init_decl->type = ResolveType(type->Clone(), std::move(init_decl->type));
@@ -405,7 +410,13 @@ decl: declaration_specifiers init_declarator_list_opt SEMICOLON {
       }
     } else {
       auto decl = std::move(std::get<std::unique_ptr<DeclNode>>(decl_specifiers));
+      // A record declaration that doesn't declare any identifier, e.g., `struct point {int x, int y};`.
+      if (init_decl_list.empty()) {
+        decl_list.push_back(std::move(decl));
+      }
+
       auto* rec_decl = dynamic_cast<RecordDeclNode*>(decl.get());
+      // Initialize record variable.
       for (auto& init_decl : init_decl_list) {
         if (init_decl) {
           init_decl->type = ResolveType(rec_decl->type->Clone(), std::move(init_decl->type));
--- a/test/typecheck/decl.exp
+++ b/test/typecheck/decl.exp
@@ -7,6 +7,7 @@ ProgramNode <1:1>
       DeclStmtNode <3:3>
         VarDeclNode <3:7> j: int
       DeclStmtNode <4:3>
+        VarDeclNode <4:3> : int
       DeclStmtNode <5:3>
         VarDeclNode <5:7> a: int
         VarDeclNode <5:10> b: int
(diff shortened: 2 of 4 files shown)


parser.y
```diff
@@ -406,6 +406,13 @@ decl: declaration_specifiers init_declarator_list_opt SEMICOLON {
     } else {
       auto decl = std::move(std::get<std::unique_ptr<DeclNode>>(decl_specifiers));
       auto* rec_decl = dynamic_cast<RecordDeclNode*>(decl.get());
+      // NOTE: This is record type declaration. If there are objects in init_decl_list,
+      // then it should be record variable declaration.
+      if (init_decl_list.empty()) {
+        decl_list.push_back(std::move(decl));
+      }
```
Lai-YT, line 413: Suggest placing this piece of code after the declaration of `rec_decl`, as `rec_decl` is not used inside it.

Lai-YT (commented):
It seems like we have the same issue regarding to unnamed primitive types.
We check whether it declares an identifier or not inside the iteration over the `init_decl_list`; however, we don't even get into the loop when it doesn't declare an identifier, leading to an empty declaration node:
https://github.com/fruits-lab/vitaminc/blob/2f6edee121e2245d891f4483d5a4c1341883c69e/parser.y#L398-L405
https://github.com/fruits-lab/vitaminc/blob/2f6edee121e2245d891f4483d5a4c1341883c69e/test/typecheck/decl.c#L4
https://github.com/fruits-lab/vitaminc/blob/2f6edee121e2245d891f4483d5a4c1341883c69e/test/typecheck/decl.exp#L9-L10

Would you mind addressing them together? Here's a possible modification:
```bison
    if (std::holds_alternative<std::unique_ptr<Type>>(decl_specifiers)) {
      auto type = std::move(std::get<std::unique_ptr<Type>>(decl_specifiers));
      if (init_decl_list.empty()) {
        // A stand-alone type that doesn't declare any identifier, e.g., `int;`.
        decl_list.push_back(std::make_unique<VarDeclNode>(Loc(@1), "", type->Clone()));
      }

      // Declare primitive identifier.
      for (auto& init_decl : init_decl_list) {
        if (init_decl) {
          init_decl->type = ResolveType(type->Clone(), std::move(init_decl->type));
        }
        decl_list.push_back(std::move(init_decl));
      }
    } else {
      auto decl = std::move(std::get<std::unique_ptr<DeclNode>>(decl_specifiers));
      // A record declaration that doesn't declare any identifier, e.g., `struct point {int x, int y};`.
      if (init_decl_list.empty()) {
        decl_list.push_back(std::move(decl));
      }

      auto* rec_decl = dynamic_cast<RecordDeclNode*>(decl.get());
      // Declare record identifier.
      for (auto& init_decl : init_decl_list) {
        if (init_decl) {
          init_decl->type = ResolveType(rec_decl->type->Clone(), std::move(init_decl->type));
        }
        decl_list.push_back(std::move(init_decl));
      }
    }
```

Lai-YT:
I just realized there is another issue. If we build a `DeclNode` with the name `""`, our code generator is likely to allocate space for it. We need to filter out these empty nodes. It is important to note that Clang doesn't build nodes for declarations that have no effect on the program.
```
`-FunctionDecl 0x1f690a0 <test/typecheck/decl.c:1:1, line:6:1> line:1:5 main 'int ()'
  `-CompoundStmt 0x1f69558 <col:12, line:6:1>
    |-DeclStmt 0x1f69228 <line:2:3, col:12>
    | `-VarDecl 0x1f691a0 <col:3, col:11> col:7 i 'int' cinit
    |   `-IntegerLiteral 0x1f69208 <col:11> 'int' 0
    |-DeclStmt 0x1f692c0 <line:3:3, col:8>
    | `-VarDecl 0x1f69258 <col:3, col:7> col:7 j 'int'
    `-DeclStmt 0x1f69540 <line:5:3, col:25>
      |-VarDecl 0x1f692f0 <col:3, col:7> col:7 a 'int'
      |-VarDecl 0x1f69370 <col:3, col:14> col:10 b 'int' cinit
      | `-IntegerLiteral 0x1f693d8 <col:14> 'int' 1
      |-VarDecl 0x1f69410 <col:3, col:17> col:17 c 'int'
      `-VarDecl 0x1f69490 <col:3, col:24> col:20 d 'int' cinit
        `-IntegerLiteral 0x1f694f8 <col:24> 'int' 0
```

leewei05:
> I just realized there is another issue. If we build a `DeclNode` with the name `""`, our code generator is likely to allocate space for it. We need to filter out these empty nodes. It is important to note that Clang doesn't build nodes for declarations that have no effect on the program.
> 

Suggest fixing this in a future PR since it's related only to unnamed primitive type nodes.

Lai-YT (approved):
LGTM :smile: 
We can resolve the code generation issue _one day_. :love_you_gesture:

PR #164, opened 2024-06-20
Remove redundant id check on function call

Lai-YT:
This PR addressed the redundant id check issue mentioned in #163, as whether the id is declared or not is already checked by the visit function of `IdExprNode`.

Additionally, I renamed a parameter that had confused me. We use abbreviations of AST node types for parameters in the visit function. However, the parameter for `RecordVarDeclNode` violates this convention, which caused confusion and made it seem like I was looking at the visit function for `RecordDeclNode`.


leewei05 (approved):
🚀

PR #165, opened 2024-06-20
Record initialization with no designators

leewei05:
This PR implement record initialization with no designator code generation. Even though we support initializing multiple elements for `union`, `union` member access expression will only return the first element (Similar with gcc and clang, they only throw warning)[1].

[1] https://godbolt.org/z/57fYnn95G


src/type.cpp
```diff
@@ -158,6 +158,17 @@ std::string StructType::id() const noexcept {
...
 
+std::size_t StructType::offset(const std::string& id) const noexcept {
+  for (auto i = std::size_t{0}, e = fields_.size(); i < e; ++i) {
+    const auto& field = fields_.at(i);
+    if (field->id == id) {
+      return i * field->type->size();
+    }
+  }
+
+  return -1;
```
Lai-YT, line 169: 1. `std::size_t` is unsigned, so we cannot return -1. If we need to return a special value when the `id` is not a record member, please mention this in the function comment. Alternatively, we can throw an exception here, as the type-checking phase should have already verified whether the `id` is a record member. Missing this check here would indicate an internal error.  
2. Is the size of each field always the same? If not, we cannot directly calculate the offset by multiplying by their index. Instead, we need to add the sizes of the fields one by one.

src/qbe_ir_generator.cpp
```diff
@@ -205,11 +205,34 @@ void QbeIrGenerator::Visit(const ArrDeclNode& arr_decl) {
...
-void QbeIrGenerator::Visit(const RecordVarDeclNode& struct_def) {}
+  for (auto i = std::size_t{0}, e = record_var_decl.inits.size(); i < e; ++i) {
+    const auto& init = record_var_decl.inits.at(i);
+    init->Accept(*this);
+    const auto init_num = num_recorder.NumOfPrevExpr();
+
+    // res_addr = base_addr + offset
+    const int res_addr_num = NextLocalNum();
+    WriteInstr_("{} =l add {}, {}", FuncScopeTemp{res_addr_num},
+                FuncScopeTemp{base_addr}, i * init->type->size());
```
Lai-YT, line 231: Shouldn't we use the `init->type->offset()` here?
We might want to overload the `offset()` function to take the index of the member so that we can support the un-designating version of initialization.

src/qbe_ir_generator.cpp
```diff
@@ -205,11 +205,34 @@ void QbeIrGenerator::Visit(const ArrDeclNode& arr_decl) {
...
-void QbeIrGenerator::Visit(const FieldNode& field) {}
+void QbeIrGenerator::Visit(const RecordVarDeclNode& record_var_decl) {
+  const auto base_addr = NextLocalNum();
+  // TODO: support different data types. We have `int` type for now.
+  WriteInstr_("{} =l alloc4 {}", FuncScopeTemp{base_addr},
+              record_var_decl.type->size());
+  id_to_num[record_var_decl.id] = base_addr;
 
-void QbeIrGenerator::Visit(const RecordVarDeclNode& struct_def) {}
+  for (auto i = std::size_t{0}, e = record_var_decl.inits.size(); i < e; ++i) {
```
Lai-YT, line 223: In the case of excess elements when initializing a union, for example, `union shape s = {3, 4, 5};`, it seems this loop iterates 3 times and stores the elements at offsets 0, 4, and 8. This behavior doesn't seem correct. However, I haven't run the code to verify this. Could you please clarify how this loop handles such cases?

Lai-YT (changes requested):

src/qbe_ir_generator.cpp
leewei05: > In the case of excess elements when initializing a union, for example, `union shape s = {3, 4, 5};`, it seems this loop iterates 3 times and stores the elements at offsets 0, 4, and 8. This behavior doesn't seem correct. However, I haven't run the code to verify this. Could you please clarify how this loop handles such cases?

I think you're right. I'll need to handle union as a special case, or else compiler will store garbage value. 🗑️

src/qbe_ir_generator.cpp
leewei05: > We might want to overload the `offset()` function to take the index of the member so that we can support the un-designating version of initialization.

Can you elaborate on this? 🤔 You mean add a `index()` function to type?

src/type.cpp
leewei05: > 1. `std::size_t` is unsigned, so we cannot return -1. If we need to return a special value when the `id` is not a record member, please mention this in the function comment. Alternatively, we can throw an exception here, as the type-checking phase should have already verified whether the `id` is a record member. Missing this check here would indicate an internal error.
> 2. Is the size of each field always the same? If not, we cannot directly calculate the offset by multiplying by their index. Instead, we need to add the sizes of the fields one by one.

Great catch! I'll fix this. 🙆‍♂️

src/qbe_ir_generator.cpp
Lai-YT: Or maybe you stop iterating when the offset is greater than the size of the record?

src/qbe_ir_generator.cpp
Lai-YT: `offset(std::size_t index)` if necessary. This depends on how you revise the expression `i * init->type->size()`. I'll elaborate after seeing the change.

src/qbe_ir_generator.cpp
leewei05: > Or maybe you stop iterating when the offset is greater than the size of the record?

Let me know what you think about mine approach. Since we only update the first member for multiple initializers, abort union code generation if the loop iterates more than once.

Lai-YT: I just find out that `struct` also allows excessive elements. Let's not make `union` special. :sweat_smile:

include/type.hpp
```diff
@@ -188,6 +188,12 @@ class RecordType : public Type {
   /// @return The type id.
   virtual std::string id()  // NOLINT(readability-identifier-naming)
       const noexcept = 0;
+  /// @return The type offset in the record based on `id`.
```
Lai-YT, line 191: Please do state that this function throws under what kind of situations, and what type of exception it throws. For example,
```suggestion
  /// @return The type offset in the record based on `id`.
  /// @throw `std::runtime_error` if the `id` is not a member of the record.
```

src/type.cpp
```diff
@@ -158,6 +158,34 @@ std::string StructType::id() const noexcept {
...
+  for (auto i = std::size_t{0}, e = fields_.size(); i < e; ++i) {
+    const auto& field = fields_.at(i);
+    if (field->id == id) {
+      return offset;
+    }
+
+    offset += field->type->size();
+  }
+
+  throw "member not found in struct!";
```
Lai-YT, line 172: Consider throwing an `exception` instead of a `string`. The basic one is `std::runtime_error`.

include/type.hpp
```diff
@@ -205,6 +211,10 @@ class StructType : public RecordType {
 
   std::string id()  // NOLINT(readability-identifier-naming)
       const noexcept override;
+  std::size_t offset(  // NOLINT(readability-identifier-naming)
+      const std::string& id) const override;
+  std::size_t offset(  // NOLINT(readability-identifier-naming)
+      const std::size_t index) const override;
```
Lai-YT, line 217: I suggest dropping the `const` specifier for pass-by-value parameters from the function declaration, but keep it in the function definition. This does not apply to `const`-reference parameters, which should be kept in both places.

```suggestion
  std::size_t offset(  // NOLINT(readability-identifier-naming)
      std::size_t index) const override;
```

Note that two function declarations differing only in the constness of their pass-by-value parameters are not considered overloads, as they are indistinguishable when called. Removing `const` from the function declaration is recommended for the following reasons:

1. The caller does not need to know whether the parameter is `const` or not, as it does not affect the caller. Including `const` only clutters the function declaration.
2. You might want to change the function's implementation, which could involve removing the `const`. If `const` is included in the function declaration, you will have to update it there as well.

src/type.cpp
```diff
@@ -158,6 +158,34 @@ std::string StructType::id() const noexcept {
...
+    const auto& field = fields_.at(i);
+    if (i == index) {
+      return offset;
+    }
+
+    offset += field->type->size();
+  }
+
+  throw "member not found in struct!";
+}
```
Lai-YT, line 187: How about using `std::accumulate` to eliminate the loop and making the member check a special case? This means checking it with an additional `if` statement instead of making it the last statement. Consider the following pseudo-code:

```cpp
std::size_t StructType::offset(const std::size_t index) const {
  if (index >= fields_.size()) {
    throw ...
  }

  return std::accumulate(fields_.cbegin(), fields_.cend(),
    std::size_t{0}, [](auto&& size, auto&& field) {
      return size + field->type->size();
    });
}
```

This approach simplifies the code, clarifies the offset is the sum of sizes, and clearly handles the case where the index is out of bounds.

Lai-YT (changes requested):

leewei05:
After applying the current fix, compiler will not initialize exceeding elements.
```
// struct.c 
int main() {
  struct ss;

  struct birth {
    int date;
    int month;
    int year;
  };

  // Compiler will stop initializing elements at index 4 since we only have
  // 3 members defined in `struct birth`.
  struct birth bd1 = {3, 5, 1998, 2000, 2002};

  __builtin_print(bd1.date);
```

```
// struct.ssa
function w $main() {
@.start.1
@.body.1
        %.1 =l alloc4 12
        %.2 =w copy 3
        %.3 =l add %.1, 0
        storew %.2, %.3
        %.4 =w copy 5
        %.5 =l add %.1, 4
        storew %.4, %.5
        %.6 =w copy 1998
        %.7 =l add %.1, 8
        storew %.6, %.7
        %.8 =l copy $__builtin_print
        %.9 =w loadw %.1
        %.10 =l add %.1, 0
        %.11 =w loadw %.10
        %.12 =w call $printf(l $__builtin_print_format, w %.11)
```

src/type.cpp
```diff
@@ -158,6 +160,35 @@ std::string StructType::id() const noexcept {
...
+  throw std::runtime_error("member not found in struct!");
+}
+
+std::size_t StructType::offset(std::size_t index) const {
+  if (index >= fields_.size()) {
+    throw std::runtime_error("index out of bound!");
+  }
+
+  return std::accumulate(
+      fields_.cbegin(), std::next(fields_.cbegin(), (long)index),
```
Lai-YT, line 183: I find the _end_ too heavy, let's extract it:
```suggestion
  auto end = std::next(fields_.cbegin(), (long)index);
  return std::accumulate(
      fields_.cbegin(), end,
```

src/type.cpp
```diff
@@ -158,6 +160,35 @@ std::string StructType::id() const noexcept {
...
+
+    offset += field->type->size();
+  }
+
+  throw std::runtime_error("member not found in struct!");
+}
+
+std::size_t StructType::offset(std::size_t index) const {
+  if (index >= fields_.size()) {
+    throw std::runtime_error("index out of bound!");
```
Lai-YT, line 179:
```suggestion
    throw std::runtime_error{"index out of bound!"};
```

src/type.cpp
```diff
@@ -158,6 +160,35 @@ std::string StructType::id() const noexcept {
...
+  for (auto i = std::size_t{0}, e = fields_.size(); i < e; ++i) {
+    const auto& field = fields_.at(i);
+    if (field->id == id) {
+      return offset;
+    }
+
+    offset += field->type->size();
+  }
+
+  throw std::runtime_error("member not found in struct!");
```
Lai-YT, line 174:
```suggestion
  throw std::runtime_error{"member not found in struct!"};
```

include/type.hpp
```diff
@@ -194,6 +194,21 @@ class RecordType : public Type {
...
+      const std::string& id) const = 0;
+  /// @note Every member in union shares the same offset 0.
+  /// @return The type offset in the record based on `index`.
+  /// @throw `std::out_of_range` if the `index` is out of range.
+  virtual std::size_t OffsetOf(  // NOLINT(readability-identifier-naming)
+      std::size_t index) const = 0;
+  /// @note For union type, if the total number of members is greater than or
+  /// equal to 1, return 1. Otherwise, return 0.
+  /// @return The total number of members in the record.
+  virtual std::size_t SlotCount()  // NOLINT(readability-identifier-naming)
```
Lai-YT, line 210:
```suggestion
  /// @note For union type, there's at most one slot.
  /// @return How many members the record can hold.
  virtual std::size_t SlotCount()  // NOLINT(readability-identifier-naming)
```

Lai-YT (commented):
Additionally, the `NOLINT` can be removed after the renaming.

Lai-YT (approved):
LGTM! Masterpiece :nerd_face:

PR #167, opened 2024-06-22
Global variable declaration type check

leewei05:
- Introduce `ExternDeclNode` for global variable and function definition. 
- Since we don't support obsolete function definition, it's quite straight forward to add grammar for supporting global variables. I kept node `ProgramNode` as our root node, let me know if you think it's better to rename it to `TransUnitNode` or some other name. I prefer keeping `ProgramNode` since we are doing some internal initialization in that node. This separate our internal logic from the AST node.
- Move the logic of checking `main` function to `FuncDefNode` because ExternDeclNode can possibly be `DeclStmtNode`.

```
translation_unit : external_declaration
	| translation_unit external_declaration

external_declaration: function_definition
	| declaration
	;
```

### DIFF
--- a/include/ast.hpp
+++ b/include/ast.hpp
@@ -220,8 +220,8 @@ struct CompoundStmtNode : public StmtNode {
   std::vector<std::unique_ptr<StmtNode>> stmts;
 };
 
-/// @brief An external definition is an external declaration that is also a
-/// definition of a function(other than an inline definition) or an object.
+/// @brief An external declaration can be a definition of a function or an
+/// object.
 struct ExternDeclNode : public AstNode {
   ExternDeclNode(
       Location loc,
@@ -236,17 +236,18 @@ struct ExternDeclNode : public AstNode {
       decl;
 };
 
-/// @brief Root of the entire program.
-struct ProgramNode : public AstNode {
+/// @brief A translation unit, which the compiler handles individually,
+/// representing a high-level entity in the compilation process.
+struct TransUnitNode : public AstNode {
   /// @note vector of move-only elements are move-only
-  ProgramNode(Location loc,
-              std::vector<std::unique_ptr<ExternDeclNode>> trans_unit)
-      : AstNode{loc}, trans_unit{std::move(trans_unit)} {}
+  TransUnitNode(Location loc,
+                std::vector<std::unique_ptr<ExternDeclNode>> extern_decls)
+      : AstNode{loc}, extern_decls{std::move(extern_decls)} {}
 
   void Accept(NonModifyingVisitor&) const override;
   void Accept(ModifyingVisitor&) override;
 
-  std::vector<std::unique_ptr<ExternDeclNode>> trans_unit;
+  std::vector<std::unique_ptr<ExternDeclNode>> extern_decls;
 };
 
 struct IfStmtNode : public StmtNode {
--- a/include/ast.hpp
+++ b/include/ast.hpp
@@ -220,17 +220,33 @@ struct CompoundStmtNode : public StmtNode {
   std::vector<std::unique_ptr<StmtNode>> stmts;
 };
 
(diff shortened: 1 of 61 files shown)


src/type_checker.cpp
```diff
@@ -190,6 +186,10 @@ void TypeChecker::Visit(FuncDefNode& func_def) {
     // TODO: redefinition of function id
   }
 
+  if (func_def.id == "main") {
+    has_main_func_ = true;
+  }
```
Lai-YT, line 191: I don't recall why we were checking for the existence of the main function, but I believe this check is unnecessary now. It's perfectly acceptable for a translation unit to not have a main function. The linker will handle this and report any issues.

include/ast.hpp
```diff
@@ -220,17 +220,33 @@ struct CompoundStmtNode : public StmtNode {
   std::vector<std::unique_ptr<StmtNode>> stmts;
 };
 
+/// @brief An external definition is an external declaration that is also a
+/// definition of a function(other than an inline definition) or an object.
+struct ExternDeclNode : public AstNode {
```
Lai-YT, line 225: This comment refers to the external definition, but the node is named `ExternDeclNode`, which is confusing. I understand this sentence is from the manual, but since we are not handling inline functions, it's fine to simplify the explanation. We know that a declaration can also be a definition.

```suggestion
/// @brief An external declaration can be a definition of a function or an object.
struct ExternDeclNode : public AstNode {
```

Lai-YT (commented):
I prefer using the name `TransUnitNode` because a program can consist of multiple translation units. The compiler handles one translation unit at a time, and the linker resolves the symbols from these translation units to create the final executable program. It's fine that we perform internal initialization within this node, as a translation unit also represents a high-level entity.

```cpp
/// @brief A translation unit, which the compiler handles individually, representing a high-level entity in the compilation process.
struct TransUnitNode : public AstNode {
```

include/ast.hpp
```diff
@@ -220,17 +220,34 @@ struct CompoundStmtNode : public StmtNode {
...
+
+/// @brief A translation unit, which the compiler handles individually,
+/// representing a high-level entity in the compilation process.
+struct TransUnitNode : public AstNode {
   /// @note vector of move-only elements are move-only
-  ProgramNode(Location loc,
-              std::vector<std::unique_ptr<FuncDefNode>> func_def_list)
-      : AstNode{loc}, func_def_list{std::move(func_def_list)} {}
+  TransUnitNode(Location loc,
+                std::vector<std::unique_ptr<ExternDeclNode>> trans_unit)
```
Lai-YT, line 244: Does the name `trans_unit` seem more suitable than `extern_decl` from your perspective?

include/ast.hpp
```diff
@@ -220,17 +220,34 @@ struct CompoundStmtNode : public StmtNode {
...
+      Location loc,
+      std::variant<std::unique_ptr<FuncDefNode>, std::unique_ptr<DeclStmtNode>>
+          extern_decl)
+      : AstNode{loc}, extern_decl{std::move(extern_decl)} {}
+
+  void Accept(NonModifyingVisitor&) const override;
+  void Accept(ModifyingVisitor&) override;
+
+  std::variant<std::unique_ptr<FuncDefNode>, std::unique_ptr<DeclStmtNode>>
+      extern_decl;
```
Lai-YT, line 236: Maybe just `decl` will suffice? Since the struct is already named `ExternDeclNode`, accessing the field as `extern_decl.extern_decl` seems redundant.

```suggestion
  std::variant<std::unique_ptr<FuncDefNode>, std::unique_ptr<DeclStmtNode>> 
      decl;
```

Lai-YT (approved):
LGTM! A PR a day keeps the doctor away. :love_you_gesture: (cough)

PR #169, opened 2024-06-22
Remove tidy diff to always check the entire project

Lai-YT:
I find checking only the diff for pushes to be quite strange, as the CI might pass on the next commit even if the previous issues weren't addressed. To ensure clarity and consistency, let's always check the entire project.


leewei05 (approved):
🫡

PR #170, opened 2024-06-24
Global variable code generation

leewei05:
- Introduce `is_global` member for `DeclNode` and `ExprNode`.
- Update `is_global` in type checking stage.
- Code generation support for global variables, global arrays. I'll update global struct and union after finishing their code generation part in block scope.
- Since declaring and initializing global variables are outside of main scope, we need a solution for storing initialized values and types. Therefore, I create a new `struct GlobalVarInitVal` with members of `value` and `type` and a function `GenerateQBEInit`, so that we can generate corresponding QBE IR.

```
export data $a = align 4 { w 1, w 2, w 3, }
export data $c = align 4 { z 4 }
export
function w $main() {
@start.1
	%.5 =l alloc4 8
@body.2
	%.1 =w loadw $c
	%.2 =l extsw 0
	%.3 =l mul %.2, 4
	%.4 =l add $a, %.3
	storew 2, %.4
	storew 1, %.5
	%.6 =l add %.5, 4
	storew 1, %.6
	ret 0
}
```

P.S: I also found a bug for struct and union declaration. I'll fix that first before jumping into struct, union code generation.


Lai-YT (commented):
If I understand correctly, in the type checker, declarations and their corresponding `InitExpr` are set as global if they are in _file_ scope. Additionally, `IdExpr` and `ArrSubExpr` are set as global if they refer to global identifiers. However, since global variables can only be initialized with constant expressions (such as constant integers and enums), is it possible that we don't need to mark them specially? Constant expression nodes hold their value, so we can directly access those values when generating code, which means `GlobalVarInitVal` can be removed as well. For `ArrSubExpr`, we can check whether they refer to a global array during code generation.

With these changes, only `DeclNode` and `IdExprNode` need to hold the `is_global` data member. Additionall, we can merge the `id_is_global` map with the symbol table, making it one of its fields.

Please consider these possible modifications, and let me know if there's anything I might have misunderstood!

leewei05:
> If I understand correctly, in the type checker, declarations and their corresponding `InitExpr` are set as global if they are in _file_ scope. Additionally, `IdExpr` and `ArrSubExpr` are set as global if they refer to global identifiers. However, since global variables can only be initialized with constant expressions (such as constant integers and enums), is it possible that we don't need to mark them specially? 

Not possible with the current infrastructure. We still need to mark `DeclNode` and `ExprNode` with `is_global` data member, or else the compiler doesn't know which type of output to generate since global variable declaration and variable declaration in a block scope is completely different.

> Constant expression nodes hold their value, so we can directly access those values when generating code, which means `GlobalVarInitVal` can be removed as well. For `ArrSubExpr`, we can check whether they refer to a global array during code generation.

It is doable for `VarDeclNode` but not possible for `ArrDeclNode`.  As you can see in our AST dump output, `VarDeclNode`'s initialization node is `IntConstExprNode`, which we can use cast `decl.init` node as `IntConstExprNode` and use its constant value. However, `ArrDeclNode`'s initialization is a series of `InitExprNode`, we would need some way to get the integer value from the bottom level, unless you want every AST node to have a constant value member, which I don't prefer. And I haven't consider the case for `SimpleAssignmentExprNode` yet. Unless we can store some kind of data structure like `llvm::Value*`, I think we would still need additional data structure, such as `GlobalVarInitVal`, to implement global variable code generation.

```
       DeclStmtNode <14:3>
          VarDeclNode <14:7> b: int
            IntConstExprNode <14:11> 2: int
          VarDeclNode <14:14> c: int
            IntConstExprNode <14:18> 0: int
          ArrDeclNode <14:21> d: int[4]
            InitExprNode <14:29> int
              IntConstExprNode <14:29> 1: int
            InitExprNode <14:29> int
              IdExprNode <14:32> b: int
            InitExprNode <14:29> int
              SimpleAssignmentExprNode <14:37> int
                IdExprNode <14:35> c: int
                IntConstExprNode <14:39> 3: int
            InitExprNode <14:29> int
              IntConstExprNode <14:42> 4: int
```

> Additionally, we can merge the `id_is_global` map with the symbol table, making it one of its fields.

1. This would require some changes in `scope.cpp` and we would need to keep the global symbol table in type checking stage til the code generation stage.
2.  This approach also assumes that we can get the `id` of our expression. For example, a global pointer variable.

```
int b = 10;
int *ptr = &b;

int main() {
    *ptr;
    return 0;
}
```
There's a `UnaryExprNode` in the way of getting to `IdExprNode`, so dynamic_cast couldn't work and we cannot get the `id` to search for symbol in the symbol table.

```
TransUnitNode <1:1>
  ExternDeclNode <1:1>
    DeclStmtNode <1:1>
      VarDeclNode <1:5> b: int
        IntConstExprNode <1:9> 10: int
  ExternDeclNode <2:1>
    DeclStmtNode <2:1>
      VarDeclNode <2:6> ptr: int*
        UnaryExprNode <2:12> int* &
          IdExprNode <2:13> b: int
  ExternDeclNode <4:1>
    FuncDefNode <4:5> main: int ()
      CompoundStmtNode <4:12>
        ExprStmtNode <5:5>
          FuncCallExprNode <5:5> int
            IdExprNode <5:5> __builtin_print: int (int)
            ArgExprNode <5:21> int
              UnaryExprNode <5:21> int *
                IdExprNode <5:22> ptr: int*
        ReturnStmtNode <6:5>
          IntConstExprNode <6:12> 0: int
```

In conclusion, I agree that not every AST node needs `is_global` data member, so I can change to only `DeclNode` and `ExprNode` that have `is_global` data member.  For your other suggestions, I prefer not to change it. Since LLVM is already ported as one of our backend, I prefer to focus on that. QBE will still be supported alongside LLVM in our future feature, but I don't want to do `too much hacking` on it.

leewei05:
Btw, I haven't rebase and implemented LLVM's part yet! 🫡

src/qbe_ir_generator.cpp
```diff
@@ -644,28 +715,41 @@ void QbeIrGenerator::Visit(const IdExprNode& id_expr) {
...
-  reg_num_to_id_num[reg_num] = id_num;
 }
 
 void QbeIrGenerator::Visit(const IntConstExprNode& int_expr) {
-  int num = NextLocalNum();
-  WriteInstr_("{} =w copy {}", FuncScopeTemp{num}, int_expr.val);
-  num_recorder.Record(num);
+  if (int_expr.is_global) {
+    global_var_init_vals.push_back(
+        {int_expr.val, std::make_shared<PrimType>(PrimitiveType::kInt)});
```
Lai-YT, line 747:
```suggestion
    global_var_init_vals.emplace_back(
        int_expr.val, std::make_shared<PrimType>(PrimitiveType::kInt));
```

src/type_checker.cpp
```diff
@@ -426,6 +465,9 @@ void TypeChecker::Visit(ArrSubExprNode& arr_sub_expr) {
   assert(arr_type);
   // arr_sub_expr should have the element type of the array.
   arr_sub_expr.type = arr_type->element_type().Clone();
+  if (arr_sub_expr.arr->is_global) {
+    arr_sub_expr.is_global = true;
+  }
```
Lai-YT, line 470:
```suggestion
  arr_sub_expr.is_global = arr_sub_expr.arr->is_global;
```

Lai-YT (changes requested):
I see. :ok_man: Let's only restrict the scope of `is_global` and keep the others as-is.

src/type_checker.cpp
leewei05: Fix applied!

src/qbe_ir_generator.cpp
leewei05: It'll throw this error after changing to emplace back, I'll keep the original implementation.
```
/usr/include/c++/12/bits/new_allocator.h:175:11: error: no matching function for call to ‘{anonymous}::GlobalVarInitVal::GlobalVarInitVal(const int&, std::shared_ptr<PrimType>)’
  175 |         { ::new((void *)__p) _Up(std::forward<_Args>(__args)...); }
      |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Lai-YT: Oh, this is because `GlobalVarInitVal` only has compiler-provided constructors, which are the default constructor and the copy/move constructors. Never mind, keeping the original one is totally fine. 👍

src/qbe_ir_generator.cpp
```diff
@@ -144,65 +176,104 @@ void QbeIrGenerator::Visit(const DeclStmtNode& decl_stmt) {
...
+  if (arr_decl.is_global) {
+    const auto* arr_type = dynamic_cast<ArrType*>((arr_decl.type).get());
+    assert(arr_type);
+    if (arr_decl.init_list.size() != 0) {
+      global_var_init_vals.clear();
+      Write_("export data {} = align {} {{ ",
+             user_defined::GlobalPointer{arr_decl.id},
+             arr_type->element_type().size());
+      for (auto i = std::size_t{0}, e = arr_type->len(); i < e; ++i) {
+        auto& arr_init = arr_decl.init_list.at(i);
```
Lai-YT, line 233: Is it possible that the `init_list` is shorter than the `arr_type->len()`? If yes, those elements have to be set as 0, which can be done using `z {remaining bytes}`.

src/qbe_ir_generator.cpp
```diff
@@ -144,65 +176,104 @@ void QbeIrGenerator::Visit(const DeclStmtNode& decl_stmt) {
...
+  } else {
+    int base_addr_num = NextLocalNum();
+    assert(arr_decl.type->IsArr());
+    const auto* arr_type = dynamic_cast<ArrType*>((arr_decl.type).get());
+    auto element_size = arr_type->element_type().size();
+    WriteInstr_("{} =l alloc{} {}", FuncScopeTemp{base_addr_num}, element_size,
+                arr_decl.type->size());
+    id_to_num[arr_decl.id] = base_addr_num;
+
+    for (auto i = std::size_t{0}, e = arr_type->len(); i < e; ++i) {
```
Lai-YT, line 255: I just realized an issue here: the array should not be filled with 0s if the `init_list` is empty. Is that handled here? If not, it's probably a bug from previous PRs. Addressing it in this PR or a future PR would both be appreciated. 😊

src/llvm_ir_generator.cpp
```diff
@@ -163,39 +163,81 @@ void LLVMIRGenerator::Visit(const DeclStmtNode& decl_stmt) {
...
+    auto global_arr = module_.getOrInsertGlobal(arr_decl.id, type);
+    id_to_val[arr_decl.id] = global_arr;
+  } else {
+    auto addr = builder_.CreateAlloca(type);
+    id_to_val[arr_decl.id] = addr;
+    base_addr = addr;
+  }
 
   auto arr_decl_type = dynamic_cast<ArrType*>(arr_decl.type.get());
+  std::vector<llvm::Constant*> arr_elems{};
```
Lai-YT, line 204: Please add a brief comment stating that this vector stores the initial values of the global array.

test/codegen/global_decl.c
Lai-YT, line 1: Suggest including some value changes to the globals before printing their values. This ensures we are not placing them in read-only areas.

Lai-YT (commented):
Thank you for implementing these; this is yet another complicated PR. :sob:

src/qbe_ir_generator.cpp
leewei05: Updated!

src/qbe_ir_generator.cpp
leewei05: Since it's not possible to check the value of uninitialized array, we can check the output of both `qbe` and `llvm` for `test/codegen/array.c`.

```
int main() {
  ...
  int f[2];
  return 0;
}
```
QBE IR
```
export
function w $main() {
@.start.1
@.body.1
        %.147 =w call $printf(l $__builtin_print_format, w %.146)
	%.148 =l alloc4 8
        # no default initialization for array
	%.149 =w copy 0
	ret %.149
}
```

LLVM IR
```
define i32 @main() {
body:
  ...
  %72 = call i32 @printf(ptr @__builtin_print_format, i32 %71)
  %73 = alloca [2 x i32], align 4
  // no default initialization for array
  ret i32 0
}
```

src/llvm_ir_generator.cpp
```diff
@@ -163,39 +163,83 @@ void LLVMIRGenerator::Visit(const DeclStmtNode& decl_stmt) {
...
+            type, id_to_val.at(arr_decl.id), 0, i);
+        builder_.CreateStore(init_val, res_addr);
+      }
     } else {
       // set remaining elements as 0
       auto zero = llvm::ConstantInt::get(builder_.getInt32Ty(), 0, true);
-      builder_.CreateStore(zero, res_addr);
+      if (arr_decl.is_global) {
+        arr_elems.push_back(zero);
+      } else if (!arr_decl.is_global && 0 < init_len) {
```
Lai-YT, line 227: Since `init_len` is unsigned, use `!= 0` to better convey the condition that it should not be 0.
```suggestion
      } else if (!arr_decl.is_global && init_len != 0) {
```

src/llvm_ir_generator.cpp
```diff
@@ -163,39 +163,83 @@ void LLVMIRGenerator::Visit(const DeclStmtNode& decl_stmt) {
...
+      } else {
+        auto res_addr = builder_.CreateConstInBoundsGEP2_32(
+            type, id_to_val.at(arr_decl.id), 0, i);
+        builder_.CreateStore(init_val, res_addr);
+      }
     } else {
       // set remaining elements as 0
       auto zero = llvm::ConstantInt::get(builder_.getInt32Ty(), 0, true);
-      builder_.CreateStore(zero, res_addr);
+      if (arr_decl.is_global) {
```
Lai-YT, line 225: Suggesting moving the comment above the entire if-condition:
```suggestion
      // A global array is always initialized to 0,
      // but a local array remains uninitialized if no values are provided.
      if (arr_decl.is_global) {
```

src/qbe_ir_generator.cpp
```diff
@@ -144,65 +176,115 @@ void QbeIrGenerator::Visit(const DeclStmtNode& decl_stmt) {
...
+           user_defined::GlobalPointer{arr_decl.id},
+           arr_type->element_type().size());
+
+    global_var_init_vals.clear();
+    auto arr_size = arr_type->len();
+    auto init_len = arr_decl.init_list.size();
+    // The predicate of this loop guarantees that it doesn't go out of bound if
+    // the number of initialized elements is less than the array declaration
+    // size.
+    for (auto i = std::size_t{0}; i < arr_size && i < init_len; ++i) {
```
Lai-YT, line 237: Since `init_len` in no larger than `arr_size`, we can remove one of the predicates. After that, we can remove the comment:
```suggestion
    assert(init_len <= arr_size);
    for (auto i = std::size_t{0}; i < init_len; ++i) {
```

src/qbe_ir_generator.cpp
```diff
@@ -144,65 +176,115 @@ void QbeIrGenerator::Visit(const DeclStmtNode& decl_stmt) {
...
+                arr_decl.type->size());
+    id_to_num[arr_decl.id] = base_addr_num;
+
+    // NOTE: Compiler will not set elements to 0 if `init_len` is 0.
+    // 6.7.9 Initialization
+    // 10. If an object that has automatic storage duration is not initialized
+    // explicitly, its value is indeterminate.
+    for (auto i = std::size_t{0}, e = arr_type->len(),
+              init_len = arr_decl.init_list.size();
+         i < e && 0 < init_len; ++i) {
```
Lai-YT, line 266: Since `init_len` is an unsigned, use `!= 0` to better convey the condition that it should have at least one init value.
```suggestion
         i < e && init_len != 0; ++i) {
```

test/codegen/array.c
```diff
@@ -22,5 +22,6 @@ int main() {
   __builtin_print(e[1]);
   __builtin_print(e[2]);
 
+  int f[2];
```
Lai-YT, line 25: Add some comment stating that the local array should not be 0-initialized, and the generated IR has to be checked manually.

src/qbe_ir_generator.cpp
```diff
@@ -1084,11 +1084,24 @@ void QbeIrGenerator::Visit(const SimpleAssignmentExprNode& assign_expr) {
...
+      WriteInstr_("storel {}, {}", FuncScopeTemp{rhs_num},
+                  FuncScopeTemp{reg_num_to_id_num.at(lhs_num)});
+    }
   } else {
-    WriteInstr_("storew {}, {}", FuncScopeTemp{rhs_num},
-                FuncScopeTemp{reg_num_to_id_num.at(lhs_num)});
+    // Global array subscripting will return the target address instead of the
+    // address of `id`.
+    if (assign_expr.lhs->is_global &&
+        !dynamic_cast<ArrSubExprNode*>(assign_expr.lhs.get())) {
```
Lai-YT, line 1109: The comment refers to subscripting a global array; however, the if block is triggered when `lhs` is a global non-subscripting array. Additionally, could you provide a small example? I'm a bit confused about why `ArrSubExprNode` is mentioned here.

Lai-YT (changes requested):

src/qbe_ir_generator.cpp
leewei05: I've updated the predicate now.

```
int b = 3
int a[3] = {6, 5, 3};

int main() {
  b = 4;
  a[0] = 7;
  return 0;
}

```

In `ArrSubExprNode`, even if an array is global, the only difference for array subscripting is to get the base address of the array from `id`. The target address is still stored in a temporary register. 

```cpp
 if (arr_sub_expr.is_global) {
    const auto id = reg_num_to_id.at(reg_num);
    WriteInstr_("{} =l add {}, {}", FuncScopeTemp{res_addr_num},
                user_defined::GlobalPointer{id}, FuncScopeTemp{offset});
  } else {
    // address of the first element
    const int base_addr = reg_num_to_id_num.at(reg_num);
    WriteInstr_("{} =l add {}, {}", FuncScopeTemp{res_addr_num},
                FuncScopeTemp{base_addr}, FuncScopeTemp{offset});
  }

  // load value from res_addr
  const int res_num = NextLocalNum();
  WriteInstr_("{} =w loadw {}", FuncScopeTemp{res_num},
              FuncScopeTemp{res_addr_num});
  reg_num_to_id_num[res_num] = res_addr_num;
  num_recorder.Record(res_num);
```

`b` will get its address from `reg_num_to_id` and `a[0]` will get its address from `reg_num_to_id_num`. Sorry for the confusion! Hopefully, this version is more clearer.

```cpp
     if (assign_expr.lhs->is_global) {
      // Global array subscripting will return the target address instead of the
      // address of `id`.
      if (dynamic_cast<ArrSubExprNode*>(assign_expr.lhs.get())) {
        WriteInstr_("storew {}, {}", FuncScopeTemp{rhs_num},
                    FuncScopeTemp{reg_num_to_id_num.at(lhs_num)});
      } else {
        WriteInstr_("storew {}, {}", FuncScopeTemp{rhs_num},
                    user_defined::GlobalPointer{reg_num_to_id.at(lhs_num)});
      }
    } else {
      WriteInstr_("storew {}, {}", FuncScopeTemp{rhs_num},
                  FuncScopeTemp{reg_num_to_id_num.at(lhs_num)});
    }
```

Lai-YT: Does this kind of issue only happen with `ArrSubExprNode`? I'm concerned that dereferencing other expressions might also require special treatment.

leewei05: > Does this kind of issue only happen with `ArrSubExprNode`? I'm concerned that dereferencing or other expressions might also require special treatment.

It might, we'll fix that if we got there.

Lai-YT: :ok_hand: Leave this as-is. :love_you_gesture:

Lai-YT (approved):
LGTM :rocket:

PR #171, opened 2024-07-01
Support LLVM backend

leewei05:
- Passed all code generation tests. 🥳 

### Changes of the old infrastructure

- Testing: I use `lli` for testing generated LLVM IR code. I assumed that `lli` can link `printf` to `libc`, which is it's own c library, and it actually worked! In order to make this work, I did a little bit of hacking in `LLVMIRGenerator::FuncCallExprNode`. Whenever, there is a `__builtin_print` function being called, it will replaced it with `printf`, so that we don't need to modify our test files. 
- `Makefile`: I added a few flags with `llvm-config`.
- `main.cpp`: separated QBE and LLVM code generation process with two helper functions. Added LLVM as target option, maybe we can consider changing LLVM as default target. 🐉 
- Type: created fields accessor for record type. Added `MemberIndex` for record type for getting the index number of members. Union will always return index `0` if member is found. Lastly, changed `return_type()` of `FuncType` to support `LLVMIRUtil::GetLLVMType()`, which may also need additional help.

### Newly introduced

- `include/llvm/util.hpp, src/llvm/util.cpp`: Some utility functions for creating LLVM IR. Also, some suggestions for storing `builder_` are appreciated! 🙏 
  - `GetLLVMType` function translate our `Type` into corresponding `llvm::Type*`, which can be directly used for `CreateCall/Load/Store` etc.
- `include/llvm_ir_generator.hpp`: Initialize some LLVM IR builder objects. I saw people say that `LLVMContext` is for LLVM developers to know the details. Right now, we only need to know `LLVMContext` includes LLVM core infrastructures. `Module` is kind of like translation unit in our parser, it knows functions, global variables and symbol table, which I didn't use (Maybe it has some handy functions we can use). `IRBuilder` is the key object for using LLVM API to construct LLVM IR.
- `src/llvm_ir_generator.cpp`: The main LLVM IR code generator file. Some differences worth noticing:
  - Comparison operator, such as, greater than, equal than, are not Binary operators in LLVM. It has a different class called `llvm::CmpInst`, so the implementation is a bit different than QBE. I think they separated them because comparison may have different low-level computer architecture optimization techniques than other binary operations.
  - I stole some of the ideas from QBE generator, such as `num_recorder`, it is `val_recorder` in LLVM IR generator instead. Since `llvm::Value` is a super important class, and any class can literally be `llvm::Value` class, except for `llvm::Type`, I decided to store `llvm::Value*` in the val_recorder. Upper level node can directly use the value stored in it.
  - We use [GetElementPtr](https://llvm.org/docs/GetElementPtr.html#introduction) for locating memory address in Array, Struct, Union.
  - Each `builder_->Create...` will create a dest number that we can use.
  - Almost every `builder_->CreateAlloc/CreateCall/CreateStore/CreateLoad` requires the callee's type. It can be tricky for function pointers because we cannot get the element type of pointer. The reason is to support [Opaque Pointers](https://llvm.org/docs/OpaquePointers.html). The solution for function pointer's type is to return `llvm::FunctionType` instead of `llvm::PointerType` for `GetLLVMType`. 
  - Every Basic block must only have one terminator instruction, such as `br`, `ret`. `llvm::BasicBlock*` can also represent as `label`. If we want to jump to label xx, we would jump to basic block xx. 

### Help wanted!

No help needed! I refactored `GetLLVMType` and remove redundant code! 😄 

~~As I mentioned that calling a function requires `llvm::FunctionType*` ( return type + parameter types), function pointers can be tricky to handle. The ideal way is to get LLVM type from our current type system, which is the job of `LLVMIRUtil::GetLLVMType`, it can return the correct LLVM type for a given Type, similar with your `ResolvedType`. Right now, I have a `val_to_type` map for mapping `llvm::Value*` to `llvm::Type*`. This way, I can get the correct function type of the corresponding `llvm::Value*` to make a function call. My plan is to get rid of `val_to_type` and fully  depend on our type system to get the correct LLVM type.~~

~~The current bottleneck is to change `PtrType`'s `base_type()` from `const Type& base_type()` to `const std::unique_ptr<Type>& base_type()`, so that I can pass the base type of `PtrType` to `LLVMIRUtil::GetLLVMType`. Changing this also require to change `IsEqual`, `IsConvertible` etc.~~


include/llvm/util.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+  /// @brief Get the corresponding LLVM type from our type.
+  llvm::Type* GetLLVMType(const std::unique_ptr<Type>& type);
+
+  LLVMIRUtil(std::unique_ptr<llvm::IRBuilder<>>& builder) : builder_{builder} {}
+
+ private:
+  /// @brief Stores a reference from the original builder.
+  std::unique_ptr<llvm::IRBuilder<>>&  // XXX: Ignore check until refactor.
+      builder_;  // NOLINT(cppcoreguidelines-avoid-const-or-ref-data-members)
```
leewei05, line 54: Let me know if you have better approach for storing this reference and not get errors from `clang-tidy`.

src/llvm/util.cpp
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  return nullptr;
+}
+
+llvm::Function* LLVMIRUtil::CurrFunc() {
+  return builder_->GetInsertBlock()->getParent();
+}
+
+llvm::Type* LLVMIRUtil::GetLLVMType(const std::unique_ptr<Type>& type) {
```
leewei05, line 79: Maybe I don't need to pass a reference of unique_ptr<Type>? 🤔 Pass `const Type&` may also work. I'll need to try this.

leewei05: Refactor done!

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+void LLVMIRGenerator::Visit(const DeclStmtNode& decl_stmt) {
+  for (const auto& decl : decl_stmt.decls) {
+    decl->Accept(*this);
+  }
+}
+
+void LLVMIRGenerator::Visit(const VarDeclNode& decl) {
+  auto var_type = llvm_util_.GetLLVMType(*(decl.type));
+  // For function pointer, we need to change from FunctionType to PointerType
+  var_type = var_type->isFunctionTy() ? var_type->getPointerTo() : var_type;
```
Lai-YT, line 159: As far as I know, the type of the `decl` itself will not be a function type. Its `init` might be a function type, but the variable itself will be a function pointer. Do we need this check here?

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+
+void LLVMIRGenerator::Visit(const FieldNode& field) {
+  /* Do nothing because this node only declares a member type in a record. */
+}
+
+void LLVMIRGenerator::Visit(const RecordVarDeclNode& record_var_decl) {
+  auto* record_type = dynamic_cast<RecordType*>(record_var_decl.type.get());
+  assert(record_type);
+  auto type = llvm_util_.GetLLVMType(*(record_var_decl.type));
+  auto base_addr = builder_->CreateAlloca(type, nullptr);
```
Lai-YT, line 207: The default value is already `nullptr`. We can omit it.
```suggestion
  auto base_addr = builder_->CreateAlloca(type);
```

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+    decl.init->Accept(*this);
+    auto val = val_recorder.ValOfPrevExpr();
+    builder_->CreateStore(val, addr);
+  }
+  id_to_val[decl.id] = addr;
+}
+
+void LLVMIRGenerator::Visit(const ArrDeclNode& arr_decl) {
+  auto arr_type = llvm_util_.GetLLVMType(*(arr_decl.type));
+  auto base_addr = builder_->CreateAlloca(arr_type, nullptr);
```
Lai-YT, line 171: The default value is already `nullptr`. We can omit it.
```suggestion
  auto base_addr = builder_->CreateAlloca(arr_type);
```

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+
+void LLVMIRGenerator::Visit(const GotoStmtNode& goto_stmt) {
+  llvm::BasicBlock* target_bb = llvm_util_.FindBBWithNameOf(goto_stmt.label);
+
+  if (target_bb) {
+    builder_->CreateBr(target_bb);
+  } else {
+    auto label_bb = llvm::BasicBlock::Create(*context_, goto_stmt.label,
+                                             llvm_util_.CurrFunc());
+    builder_->CreateBr(label_bb);
```
Lai-YT, line 409: Are we creating the basic block here if the target doesn't exist yet? If so, we can reassign the value to `target_bb` instead of declaring a new one. This better conveys the fact that we are creating one.

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+}
+
+void LLVMIRGenerator::Visit(const IdLabeledStmtNode& id_labeled_stmt) {
+  llvm::BasicBlock* target_bb =
+      llvm_util_.FindBBWithNameOf(id_labeled_stmt.label);
+
+  if (!target_bb) {
+    auto label_bb = llvm::BasicBlock::Create(*context_, id_labeled_stmt.label,
+                                             llvm_util_.CurrFunc());
+    builder_->SetInsertPoint(label_bb);
```
Lai-YT, line 465: Similarly, if we are creating the basic block here when the target doesn't exist yet, we can reassign the value to `target_bb` instead of declaring a new one. This better conveys the fact that we are creating the block.

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+}
+
+void LLVMIRGenerator::Visit(const CaseStmtNode& case_stmt) {
+  case_stmt.expr->Accept(*this);
+  auto val = val_recorder.ValOfPrevExpr();
+  auto int_expr = dynamic_cast<IntConstExprNode*>(case_stmt.expr.get());
+  assert(int_expr);
+
+  auto case_bb = llvm::BasicBlock::Create(
+      *context_, "case" + std::to_string(int_expr->val), llvm_util_.CurrFunc());
```
Lai-YT, line 480: Suggest extracting the logic of making up the label name to a variable. Additionally, maybe an underscore between the "case" and the value?
```suggestion
  auto case_label = "case_" + std::to_string(int_expr->val);
  auto case_bb = llvm::BasicBlock::Create(
      *context_, case_label, llvm_util_.CurrFunc());
```

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+
+  builder_->SetInsertPoint(case_bb);
+  case_stmt.stmt->Accept(*this);
+
+  assert(!label_views_of_jumpable_blocks.empty());
+  std::pair<llvm::Value*, llvm::BasicBlock*> p{val, case_bb};
+  label_views_of_jumpable_blocks.back().cases.push_back(p);
```
Lai-YT, line 487: This is the time to use `emplace_back`!
```suggestion
  label_views_of_jumpable_blocks.back().cases.emplace_back(val, case_bb);
```

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+
+void LLVMIRGenerator::Visit(const DefaultStmtNode& default_stmt) {
+  auto default_bb =
+      llvm::BasicBlock::Create(*context_, "default", llvm_util_.CurrFunc());
+  builder_->SetInsertPoint(default_bb);
+  default_stmt.stmt->Accept(*this);
+
+  assert(!label_views_of_jumpable_blocks.empty());
+  std::pair<llvm::Value*, llvm::BasicBlock*> p{nullptr, default_bb};
+  label_views_of_jumpable_blocks.back().cases.push_back(p);
```
Lai-YT, line 498: `emplace_back`!

Lai-YT (changes requested):
Overall, this is a great PR that makes significant progress, and I enjoyed reviewing it. As I am not familiar with the LLVM APIs, there's not much I can support. However, as all test cases have passed, I believe there are no major issues. Small ones can be found and addressed as we evolve in the future. 🔥

I have submitted some minor revisions that I hope will make the PR even better. 😄

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+      print_args.insert(print_args.end(), arg_vals.begin(), arg_vals.end());
+      builder_->CreateCall(printf, print_args);
+    } else {
+      auto called_func = module_->getFunction(func->getName());
+      auto return_res = builder_->CreateCall(called_func, arg_vals);
+      val_recorder.Record(return_res);
+    }
+  } else if (val->getType()->isPointerTy()) {
+    // function pointer
+    auto type = llvm_util_.GetLLVMType(*(call_expr.func_expr->type));
```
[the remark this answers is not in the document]
Lai-YT: When will LLVM require a function pointer to have a pointer type? I'm considering having `GetLLVMType` return the pointer type for function pointers directly. Would that be more straightforward?

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+  if (id_expr.type->IsPtr() || id_expr.type->IsFunc()) {
+    auto res = builder_->CreateLoad(llvm_util_.IntPtrType(), id_val);
+    val_recorder.Record(res);
+    val_to_id_addr[res] = id_val;
+  } else {
+    auto res =
+        builder_->CreateLoad(llvm_util_.GetLLVMType(*(id_expr.type)), id_val);
+    val_recorder.Record(res);
+    val_to_id_addr[res] = id_val;
+  }
```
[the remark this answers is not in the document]
Lai-YT: Oh I see. This is because `GetLLVMType` returns `FunctionType` on function pointers. :sweat_smile:

src/llvm_ir_generator.cpp
```diff
@@ -0,0 +1,807 @@
...
+  arr_sub_expr.arr->Accept(*this);
+  auto val = val_recorder.ValOfPrevExpr();
+  auto base_addr = val_to_id_addr.at(val);
+  auto arr_type = llvm_util_.GetLLVMType(*(arr_sub_expr.arr->type));
+  arr_sub_expr.index->Accept(*this);
+  auto index = dynamic_cast<IntConstExprNode*>(arr_sub_expr.index.get());
+  assert(index);
+
+  auto res_addr = builder_->CreateConstInBoundsGEP2_32(
+      arr_type, base_addr, 0, (unsigned int)index->val);
```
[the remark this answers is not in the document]
Lai-YT: Wow, there's a dedicated page for _GEP_. :open_mouth: It's a complex topic!

include/llvm_ir_generator.hpp
```diff
@@ -0,0 +1,87 @@
...
+  /// @brief A LLVM ostream wrapper for writing to output.
+  llvm::raw_os_ostream output_;
+  /// @brief A LLVM object that includes core LLVM infrastructure.
+  std::unique_ptr<llvm::LLVMContext> context_;
+  /// @brief Provides LLVM Builder API for constructing IR. By default, Constant
+  /// folding is enabled and we have more flexibility for inserting
+  /// instructions.
+  std::unique_ptr<llvm::IRBuilder<>> builder_;
+  /// @brief Stores global variables, function lists, and the constructed IR.
+  std::unique_ptr<llvm::Module> module_;
```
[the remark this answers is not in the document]
leewei05: Unique pointers removed!

Lai-YT (approved):
LGTM :tada:

PR #172, opened 2024-07-05
Return constant reference on query of member type

Lai-YT:
Since the `MemberType` function is a query function, it should not have to clone the type.
Note that the function is no longer `noexcept` after the change. Originally, it returned the `unknown` type when the `id` was not a member of the record. However, since we cannot make up such an unknown type as a temporary object and return its reference (undefined behavior), we have to throw an exception. This is acceptable because the type checker should already ensure that the member exists, thus querying with a non-existing `id` indicates some internal errors.


leewei05 (approved):
🚀

PR #173, opened 2024-07-06
Separate CI tests for QBE and LLVM

Lai-YT:
Currently, we have three types of tests:
- Typecheck
- QBE codegen
- LLVM codegen

These tests are run together in CI, making it difficult to identify which part is causing a failure. This PR separates these tests into individual Makefile targets so they can be run independently in CI. The behavior of `make test` remains unchanged, as it will still run all tests.


leewei05 (approved):
🚀

PR #174, opened 2024-07-06
Reduce ad-hoc type conversions on functions

Lai-YT:
By not treating function pointers specially in `GetLLVMType`, most of the ad-hoc conversions from function types to pointer types are eliminated. The only remaining conversion is when referencing the ID of a function; the type of such ID expression should be a pointer type instead of a function type. Additionally, when using such an ID expression in a function call, we need to convert it back to a function type.

### DIFF
--- a/src/llvm_ir_generator.cpp
+++ b/src/llvm_ir_generator.cpp
@@ -163,8 +163,6 @@ void LLVMIRGenerator::Visit(const DeclStmtNode& decl_stmt) {
 
 void LLVMIRGenerator::Visit(const VarDeclNode& decl) {
   auto var_type = builder_helper_.GetLLVMType(*(decl.type));
-  // For function pointer, we need to change from FunctionType to PointerType
-  var_type = var_type->isFunctionTy() ? var_type->getPointerTo() : var_type;
   auto addr = builder_.CreateAlloca(var_type);
   if (decl.init) {
     decl.init->Accept(*this);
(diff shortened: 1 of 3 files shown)


src/llvm_ir_generator.cpp
```diff
@@ -632,9 +624,18 @@ void LLVMIRGenerator::Visit(const FuncCallExprNode& call_expr) {
...
-    // function pointer
-    auto type = builder_helper_.GetLLVMType(*(call_expr.func_expr->type));
-    if (auto func_type = llvm::dyn_cast<llvm::FunctionType>(type)) {
+    llvm::Type* expr_type = nullptr;
+    if (auto ptr_type =
+            dynamic_cast<PtrType*>((call_expr.func_expr->type).get())) {
+      expr_type = builder_helper_.GetLLVMType(ptr_type->base_type());
+    } else {
+      // NOTE: Since LLVM requires the function to have pointer type when being
+      // referenced, the actual type here may still be a function type.
```
leewei05, line 633: Not sure what you meant by this. Can you give me an example? 🤔 
If `call_expr.func_expr->type->IsFunc()` is true, then it will always go to the `then` statement in the first place. The statement inside this predicate `else if (val->getType()->isPointerTy()) {` is always going to be a function pointer.

Lai-YT: Take the simplified _test/codegen/func_pointer.c_ as example:
```c
int add(int a, int b) {
  return a + b;
}

int main() {
  int (*p)(int, int) = &add;
  // The func_call.expr is `*p`, which has function type in our type system.
  // However, LLVM requires the function type to be converted to a pointer type.
  // In this point, our type system diverges from LLVM's type system:
  // val->getType()->isPointerTy() is true but call_expr.func_expr->type->IsPtr() is false.
  int d = (*p)(1, 2);
  return 0;
}
```

Lai-YT: Oops, maybe this is caused by other places. This test still fails even after I removed the conversion from function type to pointer type in the visit function of `IdExprNode`.

Lai-YT: Correction, the divergence is caused by the dereferencing of the function pointer since the LLVM type isn't actually dereferenced; it is kept as a pointer type.

leewei05 (approved):
LGTM!

PR #176, opened 2024-07-10
Use dynamic cast on reference when should not fail

Lai-YT:
There is a common pattern in our code where we dynamically cast a pointer and assert that it is not null. This pattern implies that the cast should not fail. Therefore, we can use the reference version of dynamic casting, which throws a `std::bad_cast` exception if the cast fails.
Additionally, we are using dynamic casting with unique pointers, which requires a manual `get()` to obtain the underlying pointer. Since `get()` should only be used when necessary, this change also eliminates its use.
Notice that `static_cast` has been replaced by `dynamic_cast`. This is because there is a concern that the static cast might not be valid, so we are still relying on the dynamic check.

> [!NOTE]
> The remaining use cases are those where we use dynamic checks with an `if` to verify whether an object is of a certain type, which is expected to fail.

### DIFF
--- a/parser.y
+++ b/parser.y
@@ -184,16 +184,16 @@ func_def: declaration_specifiers declarator compound_stmt {
     auto func_def = $2;
     assert(dynamic_cast<FuncDefNode*>(func_def.get()));
     assert(func_def->type->IsFunc());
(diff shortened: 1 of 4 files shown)


leewei05 (approved):
🚀

PR #177, opened 2024-07-30
Struct and Union designator initialization

leewei05:
This PR only considers test cases of identifier designator initialization.

There is an [edge case](https://godbolt.org/z/1shrvWoe3) for union initialization that I skipped for now.


leewei05:
Build failed due to https://github.com/llvm/llvm-project/issues/101201.

3 weeks ago.

```
The following additional packages will be installed:
  libclang-cpp18 libllvm18 llvm-18-dev llvm-18-linker-tools llvm-18-tools
Suggested packages:
  llvm-18-doc
The following NEW packages will be installed:
  libclang-cpp18 libllvm18 lld-18 llvm-18 llvm-18-dev llvm-18-linker-tools
  llvm-18-runtime llvm-18-tools
```

Now, with the same configurations:

```
The following additional packages will be installed:
  libllvm18 llvm-18-linker-tools
Suggested packages:
  llvm-18-doc
Recommended packages:
  llvm-18-dev
The following NEW packages will be installed:
  libllvm18 lld-18 llvm-18 llvm-18-linker-tools llvm-18-runtime
```

Manually install llvm-18-dev will have unmet dependency error

```
The following packages have unmet dependencies:
 libllvm18 : Breaks: llvm-18-dev (< 1:18.1.8~++20240730104741) but 1:18.1.8~++20240730092223+3b5b5c1ec4a3-1~exp1~20240730092238.143 is to be installed
E: Unable to correct problems, you have held broken packages.
```

leewei05:
llvm-jammy-18 distro is fixed. 😄

src/llvm_ir_generator.cpp
```diff
@@ -261,11 +267,26 @@ void LLVMIRGenerator::Visit(const RecordVarDeclNode& record_var_decl) {
   for (auto i = std::size_t{0}, e = record_var_decl.inits.size(),
             slot_count = record_type.SlotCount();
        i < slot_count && i < e; ++i) {
+    // TODO: consider the case for union when there are multiple designators. It
+    // seems that both Clang and gcc have initialized the following union with
+    // the value 6.
```
Lai-YT, line 272: I think we can directly say that the last one is used to initialized.
```suggestion
    // seems that both Clang and gcc initialized with the last one.
    // For example, following union is initialized with the value 6.
```

src/llvm_ir_generator.cpp
```diff
@@ -153,6 +153,12 @@ auto
     labels_of_jumpable_blocks  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
     = std::vector<LabelInfo>{};
 
+/// @note Designator values that allows struct and union designate the correct
+/// member for initialization.
```
Lai-YT, line 157:
```suggestion
/// @note Designator values that allows struct and union to designate the correct
/// member or index for initialization.
```

src/llvm_ir_generator.cpp
```diff
@@ -261,11 +267,26 @@ void LLVMIRGenerator::Visit(const RecordVarDeclNode& record_var_decl) {
...
     auto init_val = val_recorder.ValOfPrevExpr();
 
-    auto res_addr = builder_.CreateStructGEP(type, base_addr, i);
+    llvm::Value* res_addr = nullptr;
+    if (record_des_vals.empty()) {
+      res_addr = builder_.CreateStructGEP(type, base_addr, i);
+    } else {
+      // TODO: consider array index designators.
+      // NOTE: Only consider identifer designators for now.
+      assert(record_des_vals.size() == 1);
```
Lai-YT, line 285: Is the size of `record_des_vals` ever able to have size greater than 1? Since we always handle initialization of a single member at a time (`auto& init = record_var_decl.inits.at(i);`), I guess the answer is no?

src/llvm_ir_generator.cpp
```diff
@@ -261,11 +267,26 @@ void LLVMIRGenerator::Visit(const RecordVarDeclNode& record_var_decl) {
...
+      auto mem_index =
+          record_type.MemberIndex(std::get<std::string>(record_des_vals.at(0)));
+      res_addr = builder_.CreateStructGEP(type, base_addr, mem_index);
```
Lai-YT, line 288: The if-condition is to determine the index, thus consider moving `CreateStructGEP` outside of the loop.

```cpp
auto init_index = i;
if (!record_des_vals.empty()) {
  ...
}
res_addr = builder_.CreateStructGEP(type, base_addr, init_index);
...
```

Lai-YT (commented):
I've left some comments on the LLVM part! With minor adjustments, these suggestions can also be applied to the QBE part. :smile:

src/llvm_ir_generator.cpp
leewei05: > Is the size of `record_des_vals` ever able to have size greater than 1? 

Yes, it is possible for `record_des_vals`'s size to be greater than 1. For example,

```
struct birth {
    int date;
    int month;
    int year;
  };

  struct birth bd1[3] = {
    [2].year = 1995,
    [1].date = 3,
    [0].month = 4,
  };
```

In the above example, `bd1` is an array of `struct birth`. `[2].year` designates the last struct object's year member, which consists of a `ArrDesNode` with index 2 and a `IdDesNode` with a `year` identifier. However, I haven't considered the case with array designator. It gets more complicated with 2D or 3D array, but I don't want to get that far yet. 😕 

> Since we always handle initialization of a single member at a time (auto& init = record_var_decl.inits.at(i);), I guess the answer is no?

I think you mixed up `record_var_decl.inits.at(i)` and `record_des_vals`. In the above example, `auto& init = record_var_decl.inits.at(0);`'s `init` refers this line `[2].year = 1995`, which consists of a vector of `DesNode` and a `InitExprNode`. `record_des_vals` is the vector of `DesNode`, which has a `ArrDesNode` with index 2 and a `IdDesNode` with `year` identifier. I hope this resolves your question. 😄

Lai-YT: Oops, let me summarize to make sure I've got this right. `record_des_vals` can have a size greater than `1` if the designator is nested. In the example above, `[2].year`, `[2]` would be at index `0`, and `year` would be at index `1`. Is that correct? 😊

leewei05: Yes! That's correct!

Lai-YT (approved):
LGTM :heart_eyes:

PR #178, opened 2024-10-18
Parse record declaration while declaring variable

leewei05:
Current parser only support type declaration and variable declaration separately. With this update, users can declare both custom type and variable at the same type. Example,

```
struct animal {
  int lion;
  int tiger;
  int giraffe;
} big_zoo = {
  .lion = 10,
  .tiger = 8,
  .giraffe = 12,
};
```

Noting that I added a set for tracking custom types, so that the compiler doesn't dump type declaration every time a custom type is used.


parser.y
```diff
@@ -420,19 +424,24 @@ decl: declaration_specifiers init_declarator_list_opt SEMICOLON {
...
             init_decl->type = ResolveType(rec_decl.type->Clone(), std::move(init_decl->type));
           }
           decl_list.push_back(std::move(init_decl));
         }
       }
+
+      auto rec_type_id = dynamic_cast<RecordType&>(*rec_decl.type).id();
+      // TODO: handle cases where struct and union variables have the same name.
+      // Insert RecordDeclNode if it is a newly create type.
+      if (auto existing_type = custom_types.find(rec_type_id); existing_type == custom_types.end()) {
```
Lai-YT, line 441: I believe we can remove the temporary variable to make it an idiomatic _contains_ operation.
```suggestion
      if (custom_types.find(rec_type_id) == custom_types.end()) {
```

parser.y
```diff
@@ -420,19 +424,24 @@ decl: declaration_specifiers init_declarator_list_opt SEMICOLON {
...
+        decl_list.insert(decl_list.begin(), std::move(decl));
```
Lai-YT, line 442: Is there any special reason that the declarations have to be inserted into the beginning?

test/typecheck/union.c
```diff
@@ -26,5 +26,24 @@ int main() {
...
+    int male;
+    int female;
+  } unknown_sex;
+
+  union gender {
+    int male;
+    int female;
+  } girl = {
+    .female = 1,
+  };
```
Lai-YT, line 39: I guess the redefinition here is the _TODO_ you mentioned above, am I correct?

Lai-YT (changes requested):
COOL :sunglasses:

parser.y
leewei05: The reason is to dump type declarations in front of variable declarations.

```
          RecordDeclNode <46:10> struct book definition
            FieldNode <47:9> fiction: int
            FieldNode <48:9> sci_fi: int
            FieldNode <49:9> history: int
          RecordVarDeclNode <50:5> library: struct book
            InitExprNode <51:5> int
              IdDesNode <51:6> fiction
              IntConstExprNode <51:16> 100: int
            InitExprNode <51:5> int
              IdDesNode <52:6> sci_fi
              IntConstExprNode <52:15> 50: int
            InitExprNode <51:5> int
              IdDesNode <53:6> history
              IntConstExprNode <53:16> 500: int
```

And since we need `decl` to resolve type , I decided to insert type declaration after resolving type.

test/typecheck/union.c
leewei05: I was thinking that the TODO would fix this example, but after checking Clang https://gcc.godbolt.org/z/scj7fzsd8. It shouldn't be a correct definition. 😆 I'll remove that TODO comment.
```
struct gender {
   int male;
   int female;
};

union gender {
  int male;
  int female;
};
```

parser.y
leewei05: Fixed!

parser.y
```diff
@@ -34,6 +34,7 @@ std::unique_ptr<Type> ResolveType(std::unique_ptr<Type> resolved_type,
   #include <string>
   #include <variant>
   #include <vector>
+  #include <set>
```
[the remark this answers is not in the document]
leewei05: Fixed!

leewei05:
Added code generation tests since this doesn't require additional code generation implementations.

parser.y
Lai-YT: Could you add a comment for this? :raising_hand:

test/typecheck/union.c
Lai-YT: I see, then this is a separate issue where the test case itself redefines `gender` (see: <https://godbolt.org/z/v8Paxcqhn>). Even when a custom type is defined alongside a declaration, it introduces a new identifier, which cannot be redefined.

Lai-YT (changes requested):

test/typecheck/union.c
leewei05: > I see, then this is a separate issue where the test case itself redefines `gender` (see: https://godbolt.org/z/v8Paxcqhn). Even when a custom type is defined alongside a declaration, it introduces a new identifier, which cannot be redefined.

You're right! I've removed the redefined part in the test.

Lai-YT (approved):
LGTM!

PR #180, opened 2025-02-28
Support enum

### LINKED ISSUE #179: Support enum

Lai-YT:
This PR introduces `enum`, which acts as a scoped named constant in C.

No rush on reviewing this PR—take your time! I really appreciate you taking the time to review it.

### Type Checking

With `enum`, the concept of constant expressions must also be defined.  
Ideally, a `ConstExprNode` would wrap an `IdExprNode` representing an enumeration constant. However, to keep the parser simple without maintaining symbol tables, it cannot determine at parse time whether an identifier refers to an enumeration constant.  
Moreover, modifying the AST to insert a new node above `IdExprNode` is incompatible with our visitor pattern.

To resolve this, `IdExprNode` now holds a `ConstExprNode`, a newly introduced node currently represented only by `IntConstExprNode`.  
The type checker identifies enumeration constants by consulting the symbol table. If an identifier refers to an enumeration constant, it constructs a `ConstExprNode` with its corresponding value.

> [!IMPORTANT]  
> The introduction of constant expressions is a significant change, impacting constructs that require constant expressions but were not previously implemented due to their absence, such as `switch` cases and array initializers.
> If you have any feedback on this approach, please let me know! Support for these constructs can be added in later commits.

### Code Generation

Since `enum` values are effectively named constants, their code generation is identical to that of integer constants, making code generation as simple as delegating it to constant expression handling.  

> [!NOTE]  
> The terms *enumeration constant* and *enumerator* are interchangeable, both referring to the symbols of an enum type.  

Resolves: #179


Lai-YT:
Additionally, while implementing `enum`, I came across C’s concept of `tag` for `struct`, `union`, and `enum` as described in *6.7.2.3 Tags*.

In short, the specifier (`struct`, `union`, or `enum`) does not affect redefinition rules, meaning the following is semantically ill-formed:

```c
struct Coordinate { int x, y; };
enum   Coordinate { X, Y };
```

The identifier alone must uniquely define the type. This was previously discussed in https://github.com/fruits-lab/vitaminc/pull/163, where we initially decided to allow it by mangling type names.

I can remove the mangling in a separate PR. 😄 What do you think?

include/ast.hpp
```diff
@@ -454,17 +481,32 @@ struct NullExprNode : public ExprNode {
...
+};
+
 struct IdExprNode : public ExprNode {
   IdExprNode(Location loc, std::string id) : ExprNode{loc}, id{std::move(id)} {}
 
   void Accept(NonModifyingVisitor&) const override;
   void Accept(ModifyingVisitor&) override;
 
   std::string id;
+  /// @brief No-null if is a constant expression such as an enumeration
```
leewei05, line 503:
```suggestion
  /// @brief Not null if it is a constant expression such as an enumeration
```

test/codegen/enum.c
```diff
@@ -0,0 +1,43 @@
+enum {
+  TRUE,
+  FALSE,
+};
+
+int main() {
+  // Convertible to int.
+  int f = FALSE;
+  __builtin_print(f);
```
leewei05, line 9: nit: Since f is declared from FALSE, I assume it was 0 at first. But it is 1 due to the sequence of enum. 🤣 Consider swapping the order of TRUE, FALSE.

parser.y
```diff
@@ -423,25 +426,47 @@ decl: declaration_specifiers init_declarator_list_opt SEMICOLON {
...
-        // insert decl at the beginning of the vector.
-        decl_list.insert(decl_list.begin(), std::move(decl));
-        custom_types.insert(rec_type_id);
+        auto enum_type_id = dynamic_cast<EnumType&>(*enum_decl->type).id();
+        // Insert EnumDeclNode if it is a newly create type.
+        if (custom_types.find(enum_type_id) == custom_types.end()) {
+          // To dump type declarations in front of variable declarations,
+          // insert decl at the beginning of the vector.
+          decl_list.insert(decl_list.begin(), std::move(decl));
+          // Unnamed enums merely introduce enumeration constants.
```
leewei05, line 445: It's a bit misleading to have this comment here because the next if statement checks if enum is not unnamed. I would recommend moving this to the end of the if statement or rewriting this comment with named enums.

include/ast.hpp
```diff
@@ -175,6 +175,33 @@ struct RecordVarDeclNode : public DeclNode {
...
+  EnumConstDeclNode(Location loc, std::string id,
+                    std::unique_ptr<IntConstExprNode> value = {})
+      : DeclNode{loc, std::move(id),
+                 std::make_unique<PrimType>(PrimitiveType::kInt)},
+        value{std::move(value)} {}
+
+  void Accept(NonModifyingVisitor&) const override;
+  void Accept(ModifyingVisitor&) override;
+
+  std::unique_ptr<IntConstExprNode> value;
```
leewei05, line 189:
```suggestion
  std::unique_ptr<IntConstExprNode> int_const;
```
Better to understand with the following in the parser.
```
enum_const->int_const->val
```

parser.y
```diff
@@ -598,6 +623,44 @@ struct_or_union: STRUCT {
   }
   ;
 
+enum_specifier: ENUM id_opt LEFT_CURLY enumerator_list comma_opt RIGHT_CURLY {
+    auto decl_id = $2;
+    auto enum_list = $4;
+    auto enums = std::vector<std::unique_ptr<OptEnumConst>>{};
+    for (const auto& enum_const : enum_list) {
+      enums.push_back(std::make_unique<OptEnumConst>(enum_const->id, enum_const->value ? std::optional{enum_const->value->val} : std::nullopt));
```
leewei05, line 631:
```suggestion
      enums.push_back(std::make_unique<OptEnumConst>(enum_const->id, enum_const->value ? std::optional{enum_const->int_const->val : std::nullopt));
```

src/type_checker.cpp
```diff
@@ -442,6 +478,20 @@ void TypeChecker::Visit(IdExprNode& id_expr) {
     if (id_is_global.count(id_expr.id) != 0) {
       id_expr.is_global = true;
     }
+    // This identifier is one of the following:
+    // 1. A enumeration constant.
```
leewei05, line 482:
```suggestion
    // 1. An enumeration constant.
```

src/type_checker.cpp
```diff
@@ -442,6 +478,20 @@ void TypeChecker::Visit(IdExprNode& id_expr) {
...
+    // 2. A variable.
+    if (id_expr.type->IsEnum()) {
+      const auto& enum_type = dynamic_cast<EnumType&>(*id_expr.type);
+      // 1. Is a enumeration constant. Have it hold the value of the constant.
```
leewei05, line 486:
```suggestion
      // 1. Hold the value of the constant if it is an enumeration constant.
```

src/type_checker.cpp
```diff
@@ -442,6 +478,20 @@ void TypeChecker::Visit(IdExprNode& id_expr) {
...
+      if (enum_type.IsEnumConst(id_expr.id)) {
+        id_expr.const_expr = std::make_unique<IntConstExprNode>(
+            id_expr.loc, enum_type.ValueOf(id_expr.id));
+        id_expr.const_expr->Accept(*this);
+      } else {
+        // 2. Is a variable. Do nothing.
```
leewei05, line 492:
```suggestion
        // 2. Do nothing if it is a variable.
```

include/type.hpp
```diff
@@ -278,4 +282,56 @@ class UnionType : public RecordType {
   std::vector<std::unique_ptr<Field>> fields_;
 };
 
+/// @brief A enumerator of an enumeration, i.e., an enumeration constant.
```
leewei05, line 285:
```suggestion
/// @brief An enumerator of an enumeration, i.e., an enumeration constant.
```

include/type.hpp
```diff
@@ -278,4 +282,56 @@ class UnionType : public RecordType {
...
+struct EnumConst {
+  std::string id;
+  int value;
+
+  EnumConst(std::string id, int value) : id{std::move(id)}, value{value} {}
+};
+
+/// @brief A enumeration constant that may or may not have an explicit value.
```
leewei05, line 293:
```suggestion
/// @brief An enumeration constant may or may not have an explicit value.
```

leewei05 (changes requested):
Sorry for taking so long to review. Great work! I had some small suggestions. 😄

leewei05:
> Additionally, while implementing `enum`, I came across C’s concept of `tag` for `struct`, `union`, and `enum` as described in _6.7.2.3 Tags_.
> 
> In short, the specifier (`struct`, `union`, or `enum`) does not affect redefinition rules, meaning the following is semantically ill-formed:
> 
> ```c
> struct Coordinate { int x, y; };
> enum   Coordinate { X, Y };
> ```
> 
> The identifier alone must uniquely define the type. This was previously discussed in #163, where we initially decided to allow it by mangling type names.
> 
> I can remove the mangling in a separate PR. 😄 What do you think?

That makes sense! Please do!

parser.y
Lai-YT: I rewrote the comment! Please let me know if it's still unclear :pray:

test/codegen/enum.c
Lai-YT: I swapped the order and initialized `f` with `TRUE` instead, so that the expected output isn't changed. :love_you_gesture:

leewei05 (approved):
LGTM! Nice work! 🚀

Lai-YT:
> > Additionally, while implementing `enum`, I came across C’s concept of `tag` for `struct`, `union`, and `enum` as described in _6.7.2.3 Tags_.
> > In short, the specifier (`struct`, `union`, or `enum`) does not affect redefinition rules, meaning the following is semantically ill-formed:
> > ```c
> > struct Coordinate { int x, y; };
> > enum   Coordinate { X, Y };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The identifier alone must uniquely define the type. This was previously discussed in #163, where we initially decided to allow it by mangling type names.
> > I can remove the mangling in a separate PR. 😄 What do you think?
> 
> That makes sense! Please do!

Addressed in #181. :love_you_gesture:

PR #181, opened 2025-06-17
Drop support for struct, union, and enum sharing the same tag

Lai-YT:
Previously, we supported `struct`, `union`, and `enum` types using the same tag (identifier) by disambiguating them via name mangling in the symbol table. However, this behavior is non-standard.

According to C99 (Section 6.7.2.3, Tags):
> All declarations of structure, union, or enumerated types that have the same scope and use the same tag declare the same type.

To conform to the standard, we drop this support and no longer allow struct, union, and enum to share the same tag within the same scope.

### DIFF
--- a/src/type_checker.cpp
+++ b/src/type_checker.cpp
@@ -39,20 +39,6 @@ bool IsInBodyOf(BodyType type) {
                      [type](auto&& t) { return t == type; });
 }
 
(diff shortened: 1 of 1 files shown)


leewei05:
Why close this PR? 🤔

Lai-YT:
Oops, I deleted the branch accidentally. :sweat_smile:

PR #183, opened 2025-07-10
Move IR generators to corresponding directories

Lai-YT:
Originally, all IR generators were placed in a single location, with file names prefixed by the IR name.
Now, each IR's files are grouped in their corresponding directory, and the prefixes are no longer necessary.


leewei05 (approved):
LGTM!

